PR #17, opened 2026-05-05
TensorRT 10.16 + XL

BuffMcBigHuge:
## Summary

This PR upgrades DEMON's optional TensorRT acceleration path to TensorRT 10.16.x and hardens the surrounding build/runtime flow so it can support both `acestep-v15-turbo` and experimental `acestep-v15-xl-turbo` usage.

The main outcome is a more reliable TRT workflow: engine builds now validate the local TRT/CUDA stack, record metadata, understand checkpoint-specific decoder profiles, and select the right runtime engines for the realtime motion graph demo. The PR also adds a headless benchmark tool so TensorRT changes can be measured without browser, WebSocket, or audio-device overhead.

## What Changed

- **TensorRT dependency target moved to 10.16.x.**
  - Updates the TensorRT package range from `>=10.13,<10.14` to `>=10.16,<10.17`.
  - Tightens related acceleration dependencies, including `onnx`, `polygraphy`, and `cuda-python`.
  - Adds the PyPI TensorRT source entry and the runtime packages needed by the realtime benchmark path.

- **TensorRT build flow is now environment-aware.**
  - Adds build preflight checks for the TensorRT version, CUDA package stack, PyTorch CUDA version, active GPU, and `nvidia-smi` GPU/driver data.
  - Writes engine sidecar metadata files with the build config, ONNX checksums, environment snapshot, and creation timestamp.
  - Uses metadata matching to decide whether an existing engine can be reused or should be rebuilt.
  - Adds `--force-onnx` so an operator can explicitly regenerate ONNX before rebuilding engines.

- **Decoder precision handling was expanded for ACE-Step XL.**
  - Adds decoder precision recipes such as `fp16_mixed`, `bf16`, `bf16_mixed`, and related mixed recipes.
  - Keeps the legacy 2B mixed decoder path on `fp16_mixed`.
  - Uses `bf16_mixed` automatically for XL-looking checkpoints, which is needed for the larger model's numeric range.
  - Carries the ONNX precision recipe into TRT build metadata so engines are not accidentally reused across incompatible exports.

- **TensorRT runtime execution is stricter and dtype-correct.**
  - Decoder and VAE TRT runtimes now read per-tensor dtypes from the serialized engine instead of assuming one global IO dtype.
  - Runtime buffer allocation respects `fp16`, `bf16`, `fp32`, and other engine tensor dtypes.
  - Shape binding, tensor-address binding, shape inference, and `execute_async_v3` failures now raise explicit errors instead of failing silently.
  - VAE encode/decode outputs are normalized back to `float32` where the PyTorch side expects that format.

- **Realtime streaming can run XL b1 TRT decoder profiles.**
  - The streaming path detects the TRT decoder engine's maximum batch size.
  - If the current streaming batch exceeds the engine profile, it micro-batches rows and concatenates the results.
  - This is important for XL Turbo because the registered XL decoder engines are `b1` profiles.

- **Engine path selection is now checkpoint-aware.**
  - Adds canonical TRT engine profile registration for `acestep-v15-turbo` and `acestep-v15-xl-turbo`.
  - Keeps VAE engines shared across 2B and XL checkpoints while making decoder profiles checkpoint-specific.
  - XL currently registers 60s and 120s TRT profiles.
  - Unknown decoder checkpoints now fail clearly instead of accidentally selecting a 2B decoder engine for an incompatible model.
  - Missing-engine errors now include checkpoint-specific build commands, including the XL `--batch-max 1`, `--workspace-gb 16`, `--export-locally`, and `--decoder-precision bf16_mixed` recipe.

- **Realtime motion graph backend now selects TRT engines correctly for the active checkpoint.**
  - Caps input audio based on the largest registered profile for the selected TRT checkpoint.
  - Passes the checkpoint into `available_trt_engines()` and `smallest_fitting_profile_duration_s()`.
  - Sends an explicit `unsupported_trt_checkpoint` error to the UI when a TRT checkpoint has no registered decoder profile.
  - Keeps VAE-only TRT sessions on the shared 2B VAE profile set so VAE acceleration can still be used with non-registered decoder checkpoints.

- **A headless benchmark suite was added.**
  - Adds `demos.realtime_motion_graph_web.benchmark`.
  - Mirrors the server-side inference path: config/fixture loading, metadata detection, `Session(...)`, TRT engine selection, `prepare_source`, `encode_text`, `session.stream(...)`, repeated `stream.tick(...)`, and optional VAE decode.
  - Reports setup timings, measured tick/decode/tick+decode latency, warmup counts, decode skip counts, wall time, and CUDA peak memory.
  - Can write raw samples and summary stats to JSON for future comparisons.

- **Docs and tests were expanded.**
  - Adds `docs/TRT.md` with the current TensorRT build/runtime model, storage layout, engine naming, build commands, XL notes, realtime demo usage, troubleshooting, and validation commands.
  - Updates the realtime motion graph README with XL launch instructions and benchmark usage.
  - Extends engine-profile tests for XL profile selection, XL max duration, unknown-checkpoint rejection, VAE-only shared profiles, and XL-specific missing-engine build commands.

## Benchmark Results

Benchmark shape:

- Each run completed 28 generations: 4 warmup and 24 measured.
- Decode was enabled for all measured generations: 24 decoded, 0 skipped.
- Metrics are milliseconds unless otherwise noted.
- `tick` measures the streaming generation tick, `decode` measures VAE decode, and `tick+decode` is their combined measured latency.

### ACE-Step 1.5 Turbo

| Config | Wall time | tick mean | tick p95 | decode mean | decode p95 | tick+decode mean | tick+decode p95 | CUDA peak allocated | CUDA peak reserved |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| No TRT engine | 7779.9 | 197.4 | 198.1 | 31.4 | 33.4 | 228.8 | 230.9 | 8.13 GiB | 10.64 GiB |
| VAE decode via TRT 10.13.3.9.post1 | 4880.6 | 125.7 | 129.3 | 15.8 | 16.5 | 141.5 | 147.7 | 2.66 GiB | 2.67 GiB |
| VAE decode via TRT 10.16.1.11 | 4391.0 | 111.6 | 113.7 | 16.3 | 16.7 | 128.0 | 130.1 | 2.66 GiB | 2.67 GiB |

TRT 10.16.1.11 vs no TRT engine:

- Wall time improves from 7779.9ms to 4391.0ms, about **43.6% faster**.
- Mean `tick` improves from 197.4ms to 111.6ms, about **43.5% faster**.
- Mean `decode` improves from 31.4ms to 16.3ms, about **48.1% faster**.
- Mean `tick+decode` improves from 228.8ms to 128.0ms, about **44.1% faster**.
- CUDA peak allocated memory drops from 8.13 GiB to 2.66 GiB, about **67.3% lower**.
- CUDA peak reserved memory drops from 10.64 GiB to 2.67 GiB, about **74.9% lower**.

TRT 10.16.1.11 vs TRT 10.13.3.9.post1:

- Wall time improves from 4880.6ms to 4391.0ms, about **10.0% faster**.
- Mean `tick` improves from 125.7ms to 111.6ms, about **11.2% faster**.
- Mean `tick+decode` improves from 141.5ms to 128.0ms, about **9.5% faster**.
- Mean `decode` is roughly flat, moving from 15.8ms to 16.3ms.
- CUDA peak memory is effectively unchanged at 2.66 GiB allocated / 2.67 GiB reserved.

### ACE-Step 1.5 XL Turbo (experimental)

| Config | Wall time | tick mean | tick p95 | decode mean | decode p95 | tick+decode mean | tick+decode p95 | CUDA peak allocated | CUDA peak reserved |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| No TRT engine XL | 16773.9 | 481.8 | 483.1 | 29.6 | 31.0 | 511.4 | 515.5 | 13.01 GiB | 15.52 GiB |
| VAE decode via TRT 10.16.1.11 XL | 13566.7 | 388.2 | 390.2 | 16.8 | 17.1 | 405.0 | 407.0 | 2.66 GiB | 2.67 GiB |

TRT 10.16.1.11 XL vs no TRT engine XL:

- Wall time improves from 16773.9ms to 13566.7ms, about **19.1% faster**.
- Mean `tick` improves from 481.8ms to 388.2ms, about **19.4% faster**.
- Mean `decode` improves from 29.6ms to 16.8ms, about **43.2% faster**.
- Mean `tick+decode` improves from 511.4ms to 405.0ms, about **20.8% faster**.
- CUDA peak allocated memory drops from 13.01 GiB to 2.66 GiB, about **79.6% lower**.
- CUDA peak reserved memory drops from 15.52 GiB to 2.67 GiB, about **82.8% lower**.

## Notes For Review

- TensorRT engines remain version-, CUDA-, driver-, and GPU-architecture-specific. Engines should be rebuilt after changing any of those pieces.
- XL TensorRT support is marked experimental in this PR. The registered XL decoder profiles are currently `b1` and cover 60s and 120s durations.
- VAE TRT engines are shared between 2B and XL, but decoder TRT engines are checkpoint-specific.
- The new benchmark gives us a repeatable way to compare eager, compile, mixed backend, and TensorRT configurations going forward.

## Validation

- Added/updated unit coverage in `tests/unit/test_engine_profiles.py`.
- Added manual TensorRT validation and benchmark workflows in `docs/TRT.md`.
- Benchmarked the 2B and XL realtime motion graph paths using the headless benchmark flow described above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,10 +129,14 @@ The `workflows/covers/` directory contains standalone scripts demonstrating indi
 
 ## Running with TensorRT
 
+DEMON targets TensorRT 10.16.x for optional acceleration. TensorRT plans are
+version- and GPU-architecture-specific by default, so rebuild engines after
+changing TensorRT, CUDA, driver, or the GPU used for inference.
+
 Build TRT engines:
 
 ```bash
-# Build all engines (60s + 240s, VAE + decoder, refit + non-refit)
+# Build all engines (60s + 120s + 240s, VAE + refit decoder)
 uv run python -m acestep.engine.trt.build --all
 
 # Build 60s engines only (recommended starting point)
@@ -143,6 +147,9 @@ uv run python -m acestep.engine.trt.build --all --dry-run
 
 # Force rebuild even if engines already exist (skipped by default)
 uv run python -m acestep.engine.trt.build --all --force-rebuild
+
+# Force ONNX re-export as well as engine rebuild
+uv run python -m acestep.engine.trt.build --all --duration 60 --force-rebuild --force-onnx
 
 # Only decoders or only VAE
 uv run python -m acestep.engine.trt.build --all --decoder-only
--- a/acestep/engine/diffusion.py
+++ b/acestep/engine/diffusion.py
@@ -142,16 +142,31 @@ class DiffusionEngine:
         self._trt_stream = _get_trt_stream()
         self._trt_buf_cache = {}
 
-        # Detect I/O dtypes from engine (fp16 for mixed-precision, fp32 for legacy)
+        # Detect per-tensor I/O dtypes from the engine. Strongly-typed
+        # builds may use fp16/bf16/fp32 on different inputs, and TensorRT
+        # interprets buffers by the serialized tensor dtype.
         import tensorrt as trt
         _trt_dtype_map = {
             trt.float32: torch.float32,
             trt.float16: torch.float16,
-            trt.bfloat16: torch.bfloat16,
         }
-        hs_trt_dtype = self._trt_engine.get_tensor_dtype("hidden_states")
-        self._trt_io_dtype = _trt_dtype_map.get(hs_trt_dtype, torch.float32)
-        logger.info("TRT decoder engine ready (io_dtype={})", self._trt_io_dtype)
+        if hasattr(trt, "bfloat16"):
+            _trt_dtype_map[trt.bfloat16] = torch.bfloat16
+
+        input_names = ("hidden_states", "timestep", "encoder_hidden_states", "context_latents")
+        self._trt_input_dtypes = {
+            name: _trt_dtype_map.get(self._trt_engine.get_tensor_dtype(name), torch.float32)
+            for name in input_names
+        }
+        self._trt_output_dtype = _trt_dtype_map.get(
+            self._trt_engine.get_tensor_dtype("velocity"), torch.float32
+        )
+        self._trt_io_dtype = self._trt_input_dtypes["hidden_states"]
+        logger.info(
+            "TRT decoder engine ready (input_dtypes={}, output_dtype={})",
+            self._trt_input_dtypes,
+            self._trt_output_dtype,
+        )
 
         # Try to initialize LoRA refit manager (requires REFIT-enabled engine)
         self._lora_manager = None
@@ -440,21 +455,28 @@ class DiffusionEngine:
             ctx = self._trt_ctx
             dev = hidden_states.device
             hs_shape, ts_shape, enc_shape, cl_shape = key
-            io_dtype = self._trt_io_dtype
+            in_dtypes = self._trt_input_dtypes
 
             bufs = {
-                "hidden_states": torch.empty(hs_shape, dtype=io_dtype, device=dev),
-                "timestep": torch.empty(ts_shape, dtype=torch.float32, device=dev),
-                "encoder_hidden_states": torch.empty(enc_shape, dtype=io_dtype, device=dev),
-                "context_latents": torch.empty(cl_shape, dtype=io_dtype, device=dev),
+                "hidden_states": torch.empty(hs_shape, dtype=in_dtypes["hidden_states"], device=dev),
+                "timestep": torch.empty(ts_shape, dtype=in_dtypes["timestep"], device=dev),
+                "encoder_hidden_states": torch.empty(enc_shape, dtype=in_dtypes["encoder_hidden_states"], device=dev),
+                "context_latents": torch.empty(cl_shape, dtype=in_dtypes["context_latents"], device=dev),
             }
             for name, buf in bufs.items():
-                ctx.set_input_shape(name, tuple(buf.shape))
-                ctx.set_tensor_address(name, buf.data_ptr())
+                if not ctx.set_input_shape(name, tuple(buf.shape)):
+                    raise RuntimeError(f"TRT decoder rejected input shape for {name}: {tuple(buf.shape)}")
+                if not ctx.set_tensor_address(name, buf.data_ptr()):
+                    raise RuntimeError(f"TRT decoder rejected input address for {name}")
+
+            missing = ctx.infer_shapes()
+            if missing:
+                raise RuntimeError(f"TRT decoder shapes are insufficiently specified: {missing}")
 
             out_shape = tuple(ctx.get_tensor_shape("velocity"))
-            out_buf = torch.empty(out_shape, dtype=io_dtype, device=dev)
-            ctx.set_tensor_address("velocity", out_buf.data_ptr())
+            out_buf = torch.empty(out_shape, dtype=self._trt_output_dtype, device=dev)
+            if not ctx.set_tensor_address("velocity", out_buf.data_ptr()):
+                raise RuntimeError("TRT decoder rejected output address for velocity")
 
             self._trt_buf_cache[key] = {"bufs": bufs, "output": out_buf}
             logger.info(
@@ -478,10 +500,13 @@ class DiffusionEngine:
 
         ctx = self._trt_ctx
         for name, buf in bufs.items():
-            ctx.set_tensor_address(name, buf.data_ptr())
-        ctx.set_tensor_address("velocity", entry["output"].data_ptr())
+            if not ctx.set_tensor_address(name, buf.data_ptr()):
+                raise RuntimeError(f"TRT decoder rejected input address for {name}")
+        if not ctx.set_tensor_address("velocity", entry["output"].data_ptr()):
+            raise RuntimeError("TRT decoder rejected output address for velocity")
 
-        ctx.execute_async_v3(self._trt_stream.ptr)
+        if not ctx.execute_async_v3(self._trt_stream.ptr):
+            raise RuntimeError("TRT decoder execution failed")
         self._trt_stream.synchronize()
 
         output = entry["output"]
--- a/acestep/engine/stream.py
+++ b/acestep/engine/stream.py
@@ -249,6 +249,9 @@ class StreamPipeline:
         self._trt_stream = engine._trt_stream
         self._trt_engine = engine._trt_engine
         self._trt_io_dtype = getattr(engine, '_trt_io_dtype', torch.float32)
+        self._trt_input_dtypes = getattr(engine, "_trt_input_dtypes", {})
+        self._trt_output_dtype = getattr(engine, "_trt_output_dtype", self._trt_io_dtype)
+
         # Currently-bound TRT I/O buffers (set by _ensure_trt_bufs to one
         # entry of _trt_bufs_cache). _trt_forward reads these directly.
         self._trt_bufs: Optional[dict] = None
@@ -358,6 +361,10 @@ class StreamPipeline:
         self._trt_ctx = engine._trt_ctx
         self._trt_stream = engine._trt_stream
         self._trt_io_dtype = getattr(engine, "_trt_io_dtype", torch.float32)
+        self._trt_input_dtypes = getattr(engine, "_trt_input_dtypes", {})
+        self._trt_output_dtype = getattr(
+            engine, "_trt_output_dtype", self._trt_io_dtype
+        )
         self._trt_bufs = None
         self._trt_out_buf = None
         self._trt_bufs_cache.clear()
@@ -668,14 +675,25 @@ class StreamPipeline:
         so repeated ticks with the same ``(B, T, max_L)`` shape reuse
         allocations.
 
-        The engine is built with a fixed ``batch_max`` (typically 8).
-        If ``xt_batch.shape[0]`` exceeds the engine profile's max,
-        shape binding raises inside ``_ensure_trt_bufs``. Callers can
-        hit this when ``depth × n_conds × (1 + has_cfg)`` exceeds the
-        cap — split the batch or rebuild the engine with a larger
-        profile.
+        The active ring-buffer rows are submitted as one TensorRT
+        execution. If ``B`` exceeds the loaded engine profile, TensorRT
+        rejects the shape instead of silently falling back to sequential
+        per-row dispatch.
         """
         B, T, _ = xt_batch.shape
+        if len(timestep_list) != B:
+            raise RuntimeError(
+                "TRT stream batch mismatch: "
+                f"hidden_states batch={B}, timesteps={len(timestep_list)}"
+            )
+        if enc_batch.shape[0] != B or ctx_batch.shape[0] != B:
+            raise RuntimeError(
+                "TRT stream batch mismatch: "
+                f"hidden_states={tuple(xt_batch.shape)}, "
+                f"encoder={tuple(enc_batch.shape)}, "
+                f"context={tuple(ctx_batch.shape)}"
+            )
+
         max_L = enc_batch.shape[1]
 
         self._ensure_trt_bufs(B, T, max_L)
@@ -683,7 +701,7 @@ class StreamPipeline:
         pad = T % 2 == 1
 
         # hidden_states: cast to engine I/O dtype; pad odd T with zeros.
-        xt_io = xt_batch.to(self._trt_io_dtype)
+        xt_io = xt_batch.to(bufs["hidden_states"].dtype)
         if pad:
             bufs["hidden_states"][:, :T, :].copy_(xt_io)
             bufs["hidden_states"][:, T:, :].zero_()
@@ -697,10 +715,12 @@ class StreamPipeline:
         # encoder_hidden_states: already padded to max_L + catted by
         # the caller. The engine has no ``encoder_attention_mask``
         # input; padding is handled by zero-value convention.
-        bufs["encoder_hidden_states"].copy_(enc_batch.to(self._trt_io_dtype))
+        bufs["encoder_hidden_states"].copy_(
+            enc_batch.to(bufs["encoder_hidden_states"].dtype)
+        )
 
         # context_latents: pad odd T with zeros.
-        ctx_io = ctx_batch.to(self._trt_io_dtype)
+        ctx_io = ctx_batch.to(bufs["context_latents"].dtype)
         if pad:
             bufs["context_latents"][:, :T, :].copy_(ctx_io)
             bufs["context_latents"][:, T:, :].zero_()
@@ -712,10 +732,13 @@ class StreamPipeline:
         for name, buf in bufs.items():
             if name.startswith("_"):
                 continue
-            ctx.set_tensor_address(name, buf.data_ptr())
-        ctx.set_tensor_address("velocity", self._trt_out_buf.data_ptr())
+            if not ctx.set_tensor_address(name, buf.data_ptr()):
+                raise RuntimeError(f"TRT decoder rejected input address for {name}")
+        if not ctx.set_tensor_address("velocity", self._trt_out_buf.data_ptr()):
+            raise RuntimeError("TRT decoder rejected output address for velocity")
 
-        ctx.execute_async_v3(self._trt_stream.ptr)
+        if not ctx.execute_async_v3(self._trt_stream.ptr):
+            raise RuntimeError("TRT decoder execution failed")
         self._trt_stream.synchronize()
 
         out = self._trt_out_buf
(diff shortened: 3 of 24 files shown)


ryanontheinside:
## TRT 10.16 vs 10.13 on RTX 5090 — VAE engines crash, decoder is fine

Ran a full local A/B against this branch on a 5090 and there's a blocker: the 2B VAE engines built under TRT 10.16 segfault on the first `execute_async_v3` on this hardware. The decoder is healthy and slightly faster. Likely you only tested on the local 4090 — this looks like a Blackwell-only regression.

### Hardware / stack

- RTX 5090 (Blackwell, SM 12.0)
- NVIDIA driver `591.86`, CUDA 13.1
- Windows 11, Python 3.11, torch `2.9.1+cu128`
- main @ `24ed08c` for the 10.13 baseline, this branch (`marco/feat/trt-10-16`) @ `8466d11` for 10.16

### Decoder isolation — works, ~3 % faster on 10.16

`decoder_mixed_refit_b8_60s.engine`, fp16-mixed. Warmup 5 + 30 measured iters per shape.

| Config | TRT 10.13 mean | TRT 10.16 mean | Δ |
| --- | ---: | ---: | ---: |
| B=1, T=750, enc=200 | 8.62 ms | 8.42 ms | −2.4 % |
| B=8, T=750, enc=200 | 41.91 ms | 40.77 ms | −2.7 % |
| B=8, T=1500, enc=200 | 79.94 ms | 77.53 ms | −3.0 % |

Smaller than the 11.2 % tick speedup in the PR description, but the PR's number includes more than just the decoder kernel.

**Decoder output drift** (same fp32 deterministic input, fp16 output captured on both versions):

| Config | mean abs | max abs | rel L2 |
| --- | ---: | ---: | ---: |
| B=1, T=750 | 7.4e-3 | 6.0e-2 | 5.0e-3 |
| B=8, T=750 | 8.5e-3 | 1.3e-1 | 5.7e-3 |
| B=8, T=1500 | 8.5e-3 | 1.4e-1 | 5.6e-3 |

~0.5 % relative L2 — within the noise floor of fp16 mixed-precision tactic selection. Decoder quality is fine.

### VAE isolation — broken on 10.16

Every variant of `vae_decode_fp16_60s.engine` and `vae_encode_fp16_60s.engine` under TRT 10.16 segfaults inside `execute_async_v3` on the first call. **No TRT log emitted, exit code 139, all stdout/stderr buffers empty.**

What I tried:

| Variant | Result |
| --- | --- |
| Your shipped `*.engine.trt10.16.bak` | SIGSEGV |
| Your shipped `*.engine.trt10.16-staleONNX.bak` | SIGSEGV |
| Rebuild on TRT `10.16.1.11`, ws=16 GiB | builds OK (377 MB, 69 s), runtime SIGSEGV |
| Rebuild on TRT `10.16.1.11`, ws=4 GiB | builds OK, runtime SIGSEGV |
| Rebuild on TRT `10.16.0.72`, ws=16 GiB | builds OK, runtime SIGSEGV |
| Smallest profile shape `(1, 64, 125)` | SIGSEGV |
| Torch / null / polygraphy stream | all SIGSEGV |
| Raw `tensorrt` Python API (no polygraphy) | SIGSEGV |
| `polygraphy run …` CLI | SIGSEGV (no log) |
| **fp32 build (no FP16 flag)** | **builder fails** with `Error 10: Could not find any implementation for node {ForeignNode[ONNXTRT_unsqueezeTensor...ONNXTRT_squeezeTensor_615]}` |

The fp32 builder error is the only diagnostic TRT emits. fp16 picks a different tactic that the builder accepts but the runtime can't execute on Blackwell. Decoder engine is unaffected — it's specific to the VAE 1D-conv ONNX graph.

The 10.16 VAE engine is also **~2× larger than 10.13** (377 MB vs 180 MB for vae_decode, 339 MB vs 180 MB for vae_encode), consistent with TRT 10.16 picking a different kernel layout for this graph.

**TRT 10.13 baseline VAE numbers** (for reference):

| | mean | p50 | p95 |
| --- | ---: | ---: | ---: |
| vae_decode T=750 | 30.57 ms | 30.67 | 31.16 |
| vae_decode T=1500 | 55.56 ms | 55.68 | 56.46 |
| vae_encode 30s | 31.53 ms | 31.40 | 32.33 |
| vae_encode 60s | 55.82 ms | 56.03 | 56.56 |

### End-to-end `demos/test_stream_cover_graph.py`

Same config: `acestep-v15-turbo`, 60s source, depth=8, 8 steps, DCW on, fixed seed 1528, 160 ticks.

| Run | Wall | tick avg | vae_decode avg | per-gen | Notes |
| --- | ---: | ---: | ---: | ---: | --- |
| **TRT 10.13 — full TRT** | **17.5 s** | **79.3 ms** | **54.9 ms** | **109.6 ms** | reference |
| TRT 10.16 — full TRT | — | — | — | — | **SIGSEGV at first VAE encode in `prepare_source`** — zero ticks generated |
| TRT 10.16 — TRT decoder + eager VAE | 28.1 s | 76.3 ms | 197.5 ms | 175.2 ms | only achievable variant on 10.16, **60 % slower than 10.13** |

So the user-visible behavior on a 5090 with this PR is either a hard crash (full TRT) or a meaningful regression (decoder TRT + eager VAE).

### Hypothesis

The PR description's benchmarks claim VAE works at TRT `10.16.1.11`, but the hardware/driver stack you tested with isn't documented. If you only tested on a 4090 (Ada, SM 8.9), TRT 10.16 may compile a different VAE tactic for 12.0 (Blackwell) that fails at runtime.

I cannot test driver 591.86 against, e.g., 600.x here, but the smoking gun is:

- 10.16 builder picks a tactic where the **fp32 path errors out at build time on Blackwell** (`ForeignNode[…unsqueezeTensor…squeezeTensor_615]`)
- the same 10.16 builder, with FP16 enabled, picks a *different* tactic that builds successfully but runtime-segfaults
- 10.13 has neither problem

That pattern is consistent with a TRT 10.16 SM 12.0 regression on the VAE graph. Worth (a) confirming on a 5090 if you have access, (b) trying a newer driver, (c) reporting the ONNX (`_onnx_vae/vae_decode/vae_decode.onnx` in this PR's layout) to NVIDIA as a TRT 10.16 + Blackwell repro.

### XL TRT — already covered, don't bother re-testing here

Don't worry about the XL TRT path in this PR. That work has already been done in [`ryanontheinside/arch/xl`](https://github.com/ryanontheinside/DEMON/tree/ryanontheinside/arch/xl) with very particular care and thorough testing — see commit [`9af4645`](https://github.com/ryanontheinside/DEMON/commit/9af4645658ce7f5e8b4bc99cdb51b26e50065a9e) ("xl-accel: TRT 10.16 upgrade, VAE Myelin workaround, XL turbo benchmarks") and the accompanying `XL_ACCEL_TRT_NOTES.md`.

That commit message also independently describes the exact failure mode this PR is hitting on a 5090:

> Pin `builder_optimization_level=1` to avoid TRT 10.15+ Myelin segfault on the oobleck VAE graph (Snake + WeightNorm convs + ConvTranspose1d) on RTX 5090. Bisected: optlvl 0/1 produce a working ~360 MB engine; optlvl 2 introduces a Myelin fusion that segfaults inside `execute_async_v3`.

So the fix for the 2B VAE crash is the same one already on that branch: pin `builder_optimization_level=1` in `vae_export.py`. The XL work in `ryanontheinside/arch/xl` is the canonical reference for the TRT 10.16 upgrade — please align this PR with it rather than reinventing the workaround.

### Recommendation

Don't merge as-is. The decoder upgrade is healthy (~3 % faster, ~0.5 % output drift, no VRAM regression), but the VAE TRT path is unusable on a 5090 here without the Myelin workaround already proven on `ryanontheinside/arch/xl`. Pull the `builder_optimization_level=1` pin from that branch into `acestep/engine/trt/vae_export.py` and we should be good.

Full artifacts (bench scripts, raw timings, repro scripts for the segfault, decoder fp16 tensor captures for diff, e2e WAVs, build logs) are saved locally if useful for follow-up. Happy to share specific files.

— Claude via remote

ryanontheinside:
All raw artifacts for the validation above are committed to [`claude/pr17-rtx5090-validation`](https://github.com/ryanontheinside/DEMON/tree/claude/pr17-rtx5090-validation) (single commit [`f5c66432`](https://github.com/ryanontheinside/DEMON/commit/f5c664329e3475e7ef558bb4a498182283127bf2)) under `demos/outputs/trt_compare/`. Contents:

- `REPORT.md` — full write-up
- `bench_trt_iso.py` — decoder + VAE isolation bench harness
- `probe_engine_profiles.py` — dumps each engine's TRT optimization profile
- `compare_outputs.py` / `compare_wavs.py` — cross-version diff tools
- `debug_vae_decode{,2,3,4}.py`, `build_vae_fp32.py` — minimal repro scripts for the 10.16 VAE segfault and the fp32 builder error
- `timings_trt10_13.json`, `timings_trt10_16_dec.json` — raw isolation-bench numbers
- `outputs_trt10_13.pt` (38 MB), `outputs_trt10_16_dec.pt` — fp16 decoder output captures from deterministic-seed inputs, for cross-version diffing
- `stream_cover_trt10_13.wav`, `stream_cover_trt10_16_decoder_only.wav` — end-to-end WAVs (full-TRT 10.13 reference vs decoder-only-TRT 10.16; 10.16 full-TRT crashes before any tick)
- `*.log` — captured stdout/stderr including the segfault traces and the successful 10.16 VAE engine build log
- `pr17_comment.md` — the comment posted upthread, kept here for completeness

— Claude via remote

BuffMcBigHuge:
Update: I pulled the Blackwell/RTX 5090 lessons from the validation notes into this branch for TRT 10.16.x + ACE-Step XL support.

What changed:

- `acestep/engine/trt/vae_export.py` now pins VAE TRT builds to `builder_optimization_level = 1` and applies it to the TensorRT builder config. This is the Myelin workaround from the XL branch: TRT 10.15/10.16 default opt levels can build VAE engines that segfault on RTX 5090 during the first `execute_async_v3`.
- `acestep/engine/trt/build.py` now sets decoder `seq_opt=min(duration * 25, 1500)` when constructing `TRTBuildConfig`, so long decoder profiles such as 240s no longer inherit the unsafe 30s/750-frame opt point.
- `docs/TRT.md` now documents the RTX 5090 / Blackwell VAE caveat, the opt-level pin, and the need to rebuild VAE/DreamVAE engines on the target TRT/CUDA/driver/GPU stack.

Notes:

- I did not add a runtime eager-VAE fallback; this keeps the fix build-side as recommended and preserves full TRT acceleration.
- Existing TRT 10.16 VAE engines built before this change should be considered unsafe on Blackwell and rebuilt.
- The decoder path remains on TRT 10.16.x; the 5090 report showed decoder execution was healthy.

Validation run locally:

- `git diff --check`
- `python -m py_compile` on the touched TRT modules
- `uv run pytest tests/unit/test_engine_profiles.py -q` -> 24 passed
- lints on edited files passed

ryanontheinside:
Thanks for getting the 2B path onto 10.16. Verified on a 5090 against fresh 10.13 baselines, both runs at depth=4, steps=8, vae_window=3.0s, denoise=0.7, same fixture, `--no-fast-vae` so VAE comparison uses the canonical `vae_decode_fp16_3to30s` engine on both sides:

| Metric | 10.13 | 10.16 | Δ |
|---|---:|---:|---:|
| tick mean (DiT decoder TRT) | 44.0 ms | 42.2 ms | -4.1% |
| decode mean (VAE TRT) | 8.3 ms | 4.6 ms | -44.6% |
| tick+decode mean | 52.3 ms | 46.7 ms | -10.7% |
| Wall (28 gens) | 2960 ms | 2680 ms | -9.5% |
| CUDA peak alloc / reserved | 4.04 / 4.05 GiB | 4.04 / 4.05 GiB | unchanged |

Eager controls matched between branches within run-to-run noise; PR-side changes don't leak into the eager path. The 2B engine upgrade is solid.

The XL TRT path I'd ask you to drop from this PR. Two reasons:

**1. b1 at depth=4 defeats the ring buffer.** The whole point of the StreamDiffusion-style streaming pipeline is one batched forward pass per tick across the active slots. With a b1 engine and depth=4, `_trt_forward` micro-batches into 4 sequential single-row TRT calls per tick. That's 4× the kernel launch overhead and zero intra-batch parallelism. Measured impact: XL TRT vs eager is only -24% tick on the 5090, instead of the ~3× speedup the b4/b8 version of the same engine should produce.

**2. The b1 choice isn't actually just a VRAM call.** Tried rebuilding XL at `--batch-max 4 --workspace-gb 20` on the 5090. The engine compiled fine (84s, 7.98 GB on disk, same as b1) but the first tick at B=2 fails immediately:

```
[E] IExecutionContext::inferShapes: IShuffleLayer node_view: reshaping failed
    for tensor: linear_2
    RESHAPE input dims{2, 15360} reshape dims{1, 6, 2560}
    reshape would change volume 30720 to 15360
RuntimeError: infer_shapes error code: -1
```

The dynamo exporter bakes a literal `1` into at least one reshape op during the batch=1 example-input trace, despite `dynamic_shapes={"hidden_states": {0: Dim("batch", min=1, max=8), ...}}` being specified in `export.py:867`. Known dynamo-exporter limitation with reshape/view ops. Even with 80 GB of headroom, this ONNX cannot be served at B>1.

I already worked through the dynamic-batch ONNX patch + b8 XL build separately on `arch/xl` (walks the ONNX, replaces `[1, ...]` shape constants on Reshape nodes with `[-1, ...]`, then rebuilds), with bench data attached on that branch. I'd suggest:

- Keep all the 2B 10.16 work, the runtime per-tensor dtype fix, build metadata sidecars, and the docs.
- Drop the XL profile registration, XL build recipes, and the streaming micro-batch dispatch path. The micro-batch path is only there to support b1 XL, and b1 XL shouldn't ship.

Happy to merge once XL is removed.

BuffMcBigHuge:
Update on the latest XL feedback: I aligned this PR with the `arch/xl` dynamic-batch approach rather than shipping the b1/micro-batch workaround.

What changed:

- XL decoder profiles are now registered as true batched engines: `decoder_xl-turbo_mixed_refit_b8_60s` and `decoder_xl-turbo_mixed_refit_b8_120s`.
- The streaming TRT path no longer detects a small engine batch cap or splits active rows into sequential single-row TRT calls. It now submits the active ring-buffer rows as one batched TRT execution, so unsupported profiles fail at shape binding/inference instead of silently losing batching.
- `acestep.engine.trt.export` has the XL `bf16_mixed` export recipe restored, using dynamo export and the fp32 `proj_out` deconv island.
- `acestep.engine.trt.build` now maps XL `auto` precision to `bf16_mixed` and, for XL bf16 builds with `batch_max > 1`, rewrites the decoder ONNX before engine build. The patch walks `Reshape` shape constants that baked in `[1, ...]` and writes a sibling `*_dynbatch.onnx` with those constants changed to `[-1, ...]`, keeping it next to the source ONNX so external-data references still resolve.
- Missing-engine guidance for XL now recommends `--batch-max 8 --workspace-gb 20 --export-locally --decoder-precision bf16_mixed` instead of the old b1 recipe.
- Docs and focused profile tests were updated for the dynamic-batch XL path and the removed micro-batch dispatch behavior.

Validation run locally:

- `uv run python -m py_compile` on the touched TRT/runtime modules
- `uv run pytest tests/unit/test_engine_profiles.py -q` -> 24 passed
- lints on edited files passed

I did not run a local XL engine build in this environment, so the next validation step is to build the b8 XL engine on the target 5090/H100 stack and confirm the first B>1 tick no longer hits the `linear_2` reshape failure.

ryanontheinside (approved):
after a rebase, final cleanup of stuff we might not need, and testing from gio, im good with this!

ryanontheinside (changes requested):
Quick follow-up on the XL batched path now that 1b77dc4 has b4/b8 building cleanly via the ONNX dynbatch walk. Re-tested on the same fixture as my prior comment (xl-turbo, 5090, TRT 10.16.1.11, steps=8 depth=4 vae_window=3.0s, --no-fast-vae, 24 measured ticks) and there are two small builder knobs worth landing in this PR.

**1. Surface `--batch-opt` and `--builder-optimization-level` on the build CLI.**

`TRTBuildConfig` already carries `batch_opt: int = 1` and `builder_optimization_level: int = 3`, but neither `_run_single` nor `_run_all` exposes them, so every engine ships with the legacy `Bopt=1, opt=3` tactic set. TRT picks GEMM tactics around the OPT shape, so the XL b4/b8 engines are choosing kernels tuned for batch=1 and then running them at batch=4 in steady state.

Diff is small — two `parser.add_argument` calls in `main()` plus threading them through `_build_decoder_engine` → `TRTBuildConfig`. Defaulting both to `None` preserves current behavior for anyone not passing the flag.

**2. Build XL with `--batch-opt 4 --builder-optimization-level 5` and make b4 the canonical XL profile.**

Same machine, same fixture, mean of 24 measured ticks:

| engine | tick mean | Δ vs orig b4 |
|---|---:|---:|
| b4 (orig: bopt=1, opt=3) | 150.3 ms | — |
| b8 (orig: bopt=1, opt=3) | 148.4 ms | -1.3% |
| **b4, bopt=4, opt_level=5** | **141.7 ms** | **-5.7%** |

Build cost was 89 s vs 97 s for the original opt=3 — opt_level=5 was effectively free on this graph. The properly-tuned b4 also beats the original b8 outright, which makes sense: depth=4 streaming never exercises B>4, so the b8 engine's extra capacity is dead weight (≈8 GB on disk, larger TRT workspace at session-creation, no perf return).

Concrete asks for this PR:

- Add the two CLI flags.
- Update `_trt_build_command` to emit `--batch-max 4 --batch-opt 4 --builder-optimization-level 5` for `acestep-v15-xl-turbo` (instead of `--batch-max 8`).
- Repoint `_XL_TURBO_TRT_ENGINE_PROFILES[60.0]["decoder"]` (and 120s if it ships in this PR) at `decoder_xl-turbo_mixed_refit_b4_60s`.

Happy to send these on top of your branch if it's easier.

BuffMcBigHuge:
Implemented the requested XL builder/profile follow-up.

What changed:

- Added `--batch-opt` and `--builder-optimization-level` to `acestep.engine.trt.build`, with validation and threading into `TRTBuildConfig` for both single and `--all` decoder builds.
- Updated the XL missing-engine build command to recommend `--batch-max 4 --batch-opt 4 --builder-optimization-level 5` with the existing `--workspace-gb 20 --export-locally --decoder-precision bf16_mixed` recipe.
- Repointed the canonical XL 60s/120s profiles to `decoder_xl-turbo_mixed_refit_b4_*` and updated the focused tests, docs, and XL demo default engine name accordingly.

Validation:

- `uv run python -m py_compile acestep/engine/trt/build.py acestep/paths.py demos/test_stream_cover_graph.py`
- `uv run pytest tests/unit/test_engine_profiles.py -q` -> 24 passed
- `git diff --check`

PR #63, opened 2026-05-08
realtime_motion_graph_web: trim overlapping slice head before send

gioelecerati:
## Summary
Cuts the per-tick wire payload from the full `vae_window` (~3 s) down to roughly `advance + 25 ms` (~0.5–1 s).

## Problem
The runner re-decodes the entire `vae_window` (default 3 s, `backend.py:276`) every tick but only advances the predictive-decode head by `tick_ms + dec_ms` ≈ 0.5 s (`pipeline.py:381-390`). Each emitted slice is therefore ~80% audio the client already received — Ryan flagged it on the wire trace as 3–5 s of float16-delta payload per tick. zstd compresses the no-change overlap region OK, but param-driven micro-deltas in the overlap zone are non-zero, so we still pay 2 bytes per sample for ~80% of every slice.

## Fix
Trim the leading edge of the outgoing slice to `last_sent_end - OVERLAP_PAD` whenever the new window overlaps the previous one. The pad (1200 frames = 25 ms — same scale as the runner's seam xfade) ensures:
- any param-driven micro-deltas at the immediate seam still propagate,
- the runner's own leading-xfade landing zone (`pipeline.py:402-408`) gets one more rewrite on the next tick, so client/server divergence in that 25 ms zone is bounded to one tick (audibly negligible).

Non-contiguous windows — track loop wrap, post-swap fresh source — skip the trim and re-send the full slice from `win_start`. The `last_sent_end_ref[0] = 0` reset in the swap path makes that automatic.

## Files
- `backend.py` — three localized hunks: state init, trim before send, reset on swap.

## Test plan
- [ ] Start a session, watch wire payload sizes — should drop from ~3 s of audio per slice to ~`advance` + ~25 ms
- [ ] Audio sounds the same — no audible seam at the trim boundary
- [ ] Track loop boundary — first slice past the wrap is full-width, no audible glitch
- [ ] Mid-session source swap — first slice on the new track is full-width

## Risks
The leading 25 ms of every window (the runner's pre-xfade landing zone) is now never on the wire as a fresh value; the client retains the previous tick's values there until the next tick rewrites them. Drift is bounded to one tick. If params are changing fast and that 25 ms zone audibly diverges, increase `OVERLAP_PAD` (or set it to runner's actual xfade size if we want them to track).

### DIFF
--- a/demos/realtime_motion_graph_web/backend.py
+++ b/demos/realtime_motion_graph_web/backend.py
@@ -417,6 +417,13 @@ EAGER_MAX_PIPELINE_DEPTH = 4
 # tick. Set to 0 to disable the pause entirely (always tick).
 IDLE_PAUSE_S = float(os.environ.get("DEMON_IDLE_PAUSE_S", "20"))
 
+# Carryover kept on the wire when trimming the overlapping head of a
+# windowed slice (on_audio_ready, windowed path). Sized to match the
+# runner's leading seam xfade in pipeline.py so the xfade landing zone
+# is rewritten on the next tick and any param-driven micro-deltas at
+# the immediate seam still propagate.
+OVERLAP_PAD_FRAMES = 1200  # 25 ms at 48 kHz
+
 
 def _compute_max_pipeline_depth(diffusion_engine) -> int:
     """Largest ``pipeline_depth`` the loaded backend can serve.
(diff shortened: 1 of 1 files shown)


leszko:
Something is wrong with this code change. I tried to fix it and pushed a few more commits into it, but then there is some weird crackling heard. For now, I don't merge it.

ryanontheinside:
superseded by #177

PR #75, opened 2026-05-11
rtmg/web: tune connection-unstable pill — relax thresholds, shift above curves overlay

hthillman:
Two small fixes to the "UNSTABLE CONNECTION" pill in the rtmg web app, both tracking field feedback that it was firing too often and overlapping the curve scheduler.

## What changed

**Thresholds — relax** (`engine/networkMonitor.ts`):

| Constant | Old | New | Why |
| --- | --- | --- | --- |
| `ESCALATE_TICKS` | 6 (3s) | 12 (6s) | Above Zoom/Meet debounce range; false alarms erode trust faster than missed ones here |
| `TICK_BUDGET_RATIO` | 0.85 | 0.92 | ≤15% headroom is normal scheduler noise; trigger at ≤8% instead |
| `STALL_MS` | 1500 | 2200 | Upper end of VoIP/WebRTC convention — absorbs GC/JIT pauses |
| `RECOVERY_TICKS` | 16 | 16 | Unchanged — easy to dismiss, hard to summon |

The canvas itself visibly degrades during real trouble, so leaning conservative on the pill is the right tradeoff.

**Curve-overlay collision — shift up** (`NetworkIndicator.tsx`, `app/globals.css`):

The pill sits bottom-center; the curve scheduler's tab strip + preset menus dock against the bottom of the same viewport. They were overlapping. Subscribed to `useCurveStore.overlayOpen` and gated a `data-curves-open` attribute. CSS bumps `bottom` by `var(--curves-tab-strip-h, 64px)` when set, 240ms eased so the move is felt, not jarring.

## Test plan

- [x] HMR compiles clean
- [ ] Visual: open curve scheduler, force `quality: "unstable"` in dev tools, confirm pill clears the tab strip and transitions smoothly when overlay opens/closes
- [ ] Live: monitor unstable-fire rate on prod after merge — should drop noticeably without missing real instability events

### DIFF
--- a/demos/realtime_motion_graph_web/web/app/globals.css
+++ b/demos/realtime_motion_graph_web/web/app/globals.css
@@ -4120,6 +4120,19 @@ body[data-mode="graph"] #install-video-area .focal-side-blur { display: none; }
   animation:
     network-indicator-fade-in 200ms cubic-bezier(.2, .8, .2, 1) forwards,
     start-whisper-breathe 2.6s ease-in-out infinite 200ms;
+  transition: bottom 240ms cubic-bezier(.2, .8, .2, 1);
+}
+/* Curve scheduler overlay docks its tab strip + close button against
+   the bottom of the viewport (≈48–56px tall). When that overlay is
+   open, hoist the pill above the tabs so it doesn't sit on top of the
+   curve tabs or the preset menus that grow upward from them. */
+.network-indicator[data-curves-open="true"] {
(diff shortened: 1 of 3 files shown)


hthillman:
Superseded by a consolidated PR that bundles all four UI feedback items (network indicator, upload visibility, ribbon coil terminator, advanced-controls coachmark) on one branch so the user can test them together locally. New PR incoming.

PR #97, opened 2026-05-13
Mel-Band RoFormer Vocal Separation

BuffMcBigHuge:
Inspired by https://github.com/kijai/ComfyUI-MelBandRoFormer

DEMON-—-Realtime-Motion-Graph
DEMON-—-Realtime-Motion-Graph (1)

## Summary
- Adds Mel-Band RoFormer as the uploaded-track stem separation path for the realtime motion graph backend, with downloader/path support for `daydreamlive/MelBandRoFormer` and `MELBAND_ROFORMER_MODEL_PATH` override support.
- Extends the realtime websocket flow so custom uploads can request `full`, `vocals`, or `instruments` as the inference source while the backend streams vocal/instrumental stem assets back to the browser.
- Adds client-side stem state, source-mode selection in the upload confirmation dialog, stem overlay playback controls, and docs for the uploaded-track stem workflow.

## Behavior Notes
- Built-in fixtures keep the existing fast path and do not trigger Mel-Band extraction by default.
- Custom uploads default to `full`: inference uses the original upload, but the backend still rips vocals/instruments for optional realtime overlays.
- Selecting `vocals` or `instruments` reruns source preparation on that separated waveform so the chosen stem feeds inference.
- If stem extraction fails while `full` is still usable, the session continues and the UI receives `stem_failed`; if the selected inference source depends on a failed stem, the session/swap fails loudly.

## Validation
- `git diff --check origin/main...HEAD`
- `uv run python -m py_compile "acestep/model_downloader.py" "acestep/paths.py" "demos/realtime_motion_graph_web/backend.py" "demos/realtime_motion_graph_web/melband_reformer.py"`
- `npm run typecheck` from `demos/realtime_motion_graph_web/web`

## Remaining Risk
- Full GPU/browser E2E with an uploaded track still needs to be exercised before taking this out of draft.
- First use may download and load the RoFormer checkpoint, and uploaded-track session start/swap now includes separation latency.
- Separation quality is model-dependent: reverb, backing vocals, vocal-like synths, and doubled vocals can still leak into or out of the instrumental stem.


ryanontheinside:
Hey @BuffMcBigHuge, getting a ModuleNotFoundError: No module named 'scripts.extract_stems_melbandreformer' when running this branch.

demos/realtime_motion_graph_web/melband_reformer.py:16 (and previously backend.py) imports SAMPLE_RATE, MelBandRoformer, load_model, and separate_stems from scripts.extract_stems_melbandreformer, but that file isn't in the repo, it has no git history anywhere (git log --all -S "MelBandRoformer" only shows edits to the consumers, never an add of the script itself). Looks like it was never git add-ed. Could you push the missing scripts/extract_stems_melbandreformer.py? Thanks!

demos/realtime_motion_graph_web/backend.py
```diff
@@ -707,6 +741,43 @@ def _ms(stage: str) -> None:
             samples=int(waveform.shape[1]),
         )
     )
+    upload_stems: dict[str, torch.Tensor] | None = None
+    stem_error: str | None = None
+    if stem_source_mode is not None:
```
ryanontheinside: This block is structurally identical to the one in `apply_swap_if_pending` (~line 1884-1916): same try/extract/select-stem/re-prepare-source/error-close shape, ~35 lines of copy. Worth pulling into a helper like `_extract_and_select(waveform, session, mode)` returning `(stems, error, new_source, new_waveform)` so both call sites stay in sync.

demos/realtime_motion_graph_web/melband_reformer.py
```diff
@@ -0,0 +1,183 @@
...
+            vocals_44k, instruments_44k = separate_stems(
+                model,
+                waveform.detach().cpu().float().unsqueeze(0),
+                backend_sample_rate,
+                torch_device,
+            )
+            if torch_device.type == "cuda":
+                torch.cuda.synchronize(torch_device)
+        finally:
+            evicted = _evict_model_from_cache(torch_device)
```
ryanontheinside: The `_MODEL_CACHE` + `_MODEL_LOCK` + cache-key plumbing is dead weight as written. `extract_upload_stems` loads the model and evicts it in the same call (this finally block), so the cache is provably always empty between calls and `if evicted:` is unreachable-false. Either drop the cache and inline load+free, or stop evicting in finally and let a separate teardown release it.

demos/realtime_motion_graph_web/web/components/Performance/PerformanceShell.tsx
```diff
@@ -105,6 +109,8 @@ export function PerformanceShell() {
     <>
     <div id="performance" className="screen">
       {status === "ready" && <AudioSourceCrate />}
+      {status === "ready" && <StemOverlayPanel />}
+      <RecordButton />
```
ryanontheinside: `RecordButton` mount looks out of scope. The component existed on main but wasn't mounted; nothing in this PR's stem extraction work needs it. Split it into its own PR or call it out in the description.

demos/realtime_motion_graph_web/web/store/useCustomTracksStore.ts
```diff
@@ -28,29 +32,103 @@ interface CustomTracksState {
    *  with a File argument (the AudioSourceCrate upload path). May be
    *  empty for tracks added via other paths. */
   originalFiles: Map<string, File>;
+  /** Which version of the uploaded track should feed model inference. */
+  sourceModes: Map<string, StemSourceMode>;
```
ryanontheinside: Six parallel Maps keyed by track name (`decoded`, `originalFiles`, `sourceModes`, `stems`, `stemStatuses`, `stemErrors`) and every mutator now clones two or three of them in lockstep. A single `Map<string, CustomTrack>` with one record per name would collapse `add` / `setStems` / `setStemStatus` / `setSourceMode` into one update shape. Not strictly new (decoded/originalFiles were already parallel), but this PR makes it 3x worse.

ryanontheinside (commented):
Can you verify both flows end-to-end before merge: (1) fresh upload, pick `instruments` or `vocals` in the dialog, start session, confirm inference source is the stem (not full); (2) mid-session song swap, upload a second track with a stem mode, confirm the swap also uses the stem. Both code paths handle stems but they're duplicated, so a regression in one wouldn't necessarily show up in the other.

demos/realtime_motion_graph_web/melband_reformer.py
BuffMcBigHuge: Addressed by removing `_MODEL_CACHE`, `_MODEL_LOCK`, cache-key generation, and eviction helpers. `extract_upload_stems` now does the intended one-shot load, separate, CUDA sync, release, and cache clear inside the inference lock.

demos/realtime_motion_graph_web/web/components/Performance/PerformanceShell.tsx
BuffMcBigHuge: Addressed by removing the `RecordButton` import and mount from `PerformanceShell`. This was the result of merge with main, good catch!

demos/realtime_motion_graph_web/web/store/useCustomTracksStore.ts
BuffMcBigHuge: Addressed by replacing the parallel upload maps with `tracks: Map<string, CustomTrack>`. Each record now owns decoded audio, original file, source mode, stems, status, and error; the existing mutator API updates one cloned map.

demos/realtime_motion_graph_web/backend.py
BuffMcBigHuge: Addressed by adding `_extract_and_select_upload_stem(...)`, which returns the shared `(stems, error, source, waveform)` shape. Initial session setup and `apply_swap_if_pending` now both call it, while keeping their distinct failure responses at the call sites.

BuffMcBigHuge:
@ryanontheinside Re: the missing `scripts.extract_stems_melbandreformer` module: fixed by inlining the Mel-Band RoFormer inference/model code into `demos/realtime_motion_graph_web/melband_reformer.py` and removing the dependency on the untracked scripts file.

PR #113, opened 2026-05-19
ui(rtmg): hero-bay polish — DEATHSTEP label, Z/X kbd, dedup tooltip

hthillman:
## Summary

Three small follow-ups to #111 (now merged):

- **DEATHSTEP → Dubstep in hero bay.** `HeroStyleFader` had a local `labelFor()` that sliced the raw id, bypassing the canonical `displayLoraName` override. Route through `displayLoraName(id, name)` like every other consumer of the lora catalog.
- **Z / X keyboard shortcuts for the two hero LoRA faders.** `Z + ▲▼` nudges slot 1, `X + ▲▼` nudges slot 2. Resolves the slot's lora id at keypress time and commits via `loraStrengthDispatcher.set` so the debounce + manual-touch stamp + `sliderTargets` mirror stay coherent. Shift halves the step. Hint renders next to the fader value; honors the existing global no-kbd-hints toggle.
- **Dedup hero-bay tooltips.** `HudHelpReadout` at the top of the viewport was mirroring copy already shown by the hero-bay's own `::after` floating tooltip + kbd hint next to the value. Pass a new `excludeSelector` through `useTooltipHover` so the top-of-screen chip skips hero-bay hovers — the floating-above-the-knob tooltip stays as the single readout for those controls.

Also updates `KEY_BINDINGS` so the Config tab keyboard cheatsheet shows Z/X.

## Test plan

- [ ] Enable DEATHSTEP LoRA → hero-bay slot label reads `DUBSTEP`, not `DEATHSTEP`.
- [ ] Hold `Z` + ▲▼ → slot-1 hero fader nudges by 0.2; Shift+Z halves the step.
- [ ] Hold `X` + ▲▼ → slot-2 hero fader nudges; debounce coalesces into a single refit on release.
- [ ] Hover hero-bay knob → floating `::after` tooltip shows above the knob; HudHelpReadout chip at the top of the viewport stays hidden.
- [ ] Hover a non-hero control (e.g. AudioSourceCrate, HaloBadge) → HudHelpReadout still works.


demos/realtime_motion_graph_web/web/components/Performance/HeroMacros.tsx
```diff
@@ -109,7 +111,12 @@ function HeroStyleFader({ slotIndex }: HeroStyleFaderProps) {
   const trackRef = useRef<HTMLDivElement | null>(null);
   useLoraFaderDrag(trackRef, slotIndex, !isEmpty);
 
-  const displayLabel = loraId ? labelFor(loraId) : `Style ${slotIndex + 1}`;
+  const displayLabel = loraId
+    ? displayLoraName(loraId, catalog.find((c) => c.id === loraId)?.name)
```
ryanontheinside: The override map at `lib/loraLabels.ts:7-13` needs to go as part of this PR. Rename LoRAs in their `<stem>.metadata.json` sidecars. The `catalog.find(c => c.id === id)?.name` fallback you already wired here reads from there. Then delete `lib/loraLabels.ts` and switch the other call sites (`LibraryTile`, `MobileStepperRail`, `ScheduleCurvesOverlay`, `useEdgeLoraBinding`) to read the catalog name directly.

ryanontheinside: The override map at `lib/loraLabels.ts:7-13` needs to go as part of this PR. Rename LoRAs in their `<stem>.metadata.json` sidecars. The `catalog.find(c => c.id === id)?.name` fallback you already wired here reads from there. Then delete `lib/loraLabels.ts` and switch the other call sites (`LibraryTile`, `MobileStepperRail`, `ScheduleCurvesOverlay`, `useEdgeLoraBinding`) to read the catalog name directly.

ryanontheinside (changes requested):
Heads up: a bigger batch of LoRAs and new app-wide defaults are landing soon. Calling it out because the override-map approach this PR builds on won't scale, which is why I'm asking for it to come out as part of this PR rather than later.

ryanontheinside (changes requested):
Heads up: a bigger batch of LoRAs and new app-wide defaults are landing soon. Calling it out because the override-map approach this PR builds on won't scale, which is why I'm asking for it to come out as part of this PR rather than later.

PR #132, opened 2026-05-22
feat(rtmg): auto-reconnect WS on 1006 with slice-epoch realignment

ryanontheinside:
The WS handshake has no resume protocol — every reconnect is a full session reinit. WsReconnector runs a full-jitter backoff loop (5 attempts, 500ms → 4s, ~12s worst-case window) sized for tunnel / network blips, not pod death — recovery from a dead pod belongs to the orchestrator. The factory rebuilds RemoteBackend from cached session state and swaps the player buffer back to the clean source so the server's delta encoding (delta = generated − source) lands correctly.

player.swap() bumps swapCount, but a fresh RemoteBackend starts at _sliceEpoch=0. The slice listener's `epoch !== swapCount` guard would then drop every slice from the recovered session — symptom: clean source plays, controls dead, denoised audio never returns. setSliceEpoch(player.swapCount) closes the gap.

A failed connect's orphan ws can fire close after onGiveUp clears the reconnector; buildAndConnect's catch closes the orphan with closedByUser=true and triggerReconnect bails on terminal status as defense in depth. ResolvedFixture is snapshotted once so each retry reuses the decoded PCM instead of re-decoding the (potentially multi-MB) upload per attempt.

window.__demonDebug.simulate1006() routes a synthetic CloseEvent through the real listener chain so the recovery path is exercisable without real network failure. tests/verifyReconnect.mts covers WsReconnector retry/cancel/give-up, a real RFC6455 mid-handshake TCP destroy (close code 1006), the slice-epoch regression, and the orphan-close guard after give-up.

### DIFF
--- a/demos/realtime_motion_graph_web/web/app/RTMGBoot.tsx
+++ b/demos/realtime_motion_graph_web/web/app/RTMGBoot.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { installDemonDebug } from "@/engine/debugReconnect";
 import { listLoras } from "@/engine/lora/listLoras";
 import { setEngineUrlBuilder } from "@/engine/rtmgConfig";
 import { applyConfig, loadConfig } from "@/lib/config";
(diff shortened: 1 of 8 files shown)


gioelecerati (approved):
LGTM - needs this follow-up PR https://github.com/daydreamlive/DEMON/pull/133 to not lose timbre and structure on reconnect

PR #141, opened 2026-05-25
Added VRAM cache eviction for vae.

BuffMcBigHuge:
Solves issue of expanding VRAM leak with vae not being evicted.

### DIFF
--- a/acestep/engine/session.py
+++ b/acestep/engine/session.py
@@ -301,6 +301,12 @@ class Session:
                     " --with-dreamvae" if is_dreamvae else "",
                 )
 
+        self._trt_vae_engine_paths = tuple(
+            str(trt_engines[key])
+            for key in ("vae_encode", "vae_decode")
+            if vae_backend == "tensorrt" and key in trt_engines
+        )
+
         apply_trt_backends(
             ctx,
             decoder_backend=decoder_backend,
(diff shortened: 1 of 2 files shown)


BuffMcBigHuge:
In regards to @gioelecerati 's: https://github.com/daydreamlive/DEMON/pull/139

PR #139 targets:

- File: `acestep/engine/trt/profile_manager.py`
- Problem: stale `vae_encode_*` engines left in process-global `_trt_vae_cache`
- Timing: profile manager `bind()` and profile `_swap()`
- Goal: keep at most one encode engine resident so `_find_best_vae_engine("vae_encode")` does not pick an old 120s engine over the current 240s engine
- Scope: encode only

Our fix targets:

- File: `acestep/engine/session.py`
- Problem: `Session.close()` leaves VAE TRT engines in `_trt_vae_cache`
- Timing: session teardown, including stop/start, refresh, abandoned sessions, long-running web process cleanup
- Goal: evict this session’s `vae_encode` and `vae_decode` engines on close
- Scope: encode and decode

Important: PR #139 explicitly says decode cleanup belongs in session teardown. That is exactly what our fix does.

So the right answer is: **keep both**.

- PR #139 fixes stale encode cache correctness and mid-session/profile-swap VRAM accumulation.
- Our change fixes session teardown VRAM retention for both encode and decode.
- They touch different files and should be complementary, not competing.

gioelecerati (approved):

PR #142, opened 2026-05-25
feat(rtmg/lora): refit-aware slider — lock for ~400ms after commit

gioelecerati:
## What

After a user commits a LoRA strength change (pointer-up on the strength track), the slider **locks for 400ms** while the server-side TRT refit settles. During the lock:

- The track refuses new `pointerdown` events (`pointer-events: none` in CSS; defensively re-checked in JS).
- The thumb dims (opacity 0.55, no hover transform) and the readout dims (0.45).
- `aria-busy="true"` + a `"Applying… (LoRA refit in progress)"` tooltip explain why.

The lock auto-clears via a module-scoped `setTimeout` regardless of any server signal, so a lost packet can't soft-lock the UI.

## Why

Ryan flagged this UX in design discussion: *"the refit time is variable, and rapid updates will not behave as users expect — it is variable but certain to be not real time. the current UX is misleading."*

What actually happens today on a rapid drag:
1. `ActiveLoraRow` fires `loraStrengthDispatcher.set()` continuously during the pointer drag.
2. `useParamSync`'s 8 ms tick streams every intermediate value to the server.
3. The server runs a full TRT refit (`Refitted 192 weights in 270-295 ms` observed) **per distinct value**.

Rapid drags queue refits the server can't service in real time — sliders move but audible output lags, sometimes by seconds. The lock makes the latency honest: commit one value, see the slider lock, wait for the server to catch up, commit again.

## Scope: user gestures only

The lock is installed in the pointer-up handler. **MIDI, scheduled curves, and MCP** drivers don't go through pointer gestures and bypass `markPendingRefit` entirely — those drivers expect continuous control and there's no user holding a slider that needs to be "locked."

## Lock duration

Picked from the empirical refit cost on a fleet pod (`pod-2bde3b1d53bb`, RTX 5090). Every observed `Refitted 192 weights in NNN.N ms` sat at **265–295 ms**:

```
LoRA bach     1.400 → 0.000 : 280.5 ms
LoRA bach     0.000 → 1.400 : 285.3 ms
LoRA deathstep 1.400 → 0.000 : 271.4 ms
LoRA deathstep 0.000 → 1.400 : 295.1 ms
```

**400 ms = ~33 % headroom over the worst observed case** without feeling unresponsive on a fast refit. If profiling on larger LoRAs (384 / 576 params) shows a longer ceiling, bump `REFIT_LOCK_MS` — it's the only knob.

## Where it lands

| File | Change |
| --- | --- |
| `web/store/useLoraStore.ts` | `pendingRefit: Set<string>` state, `markPendingRefit(id)` / `clearPendingRefit(id)` actions, module-scoped `_refitTimers` map for the auto-clear, `REFIT_LOCK_MS = 400` constant. `reset()` also clears pending so import / session reset doesn't leave stale locks. |
| `web/components/Performance/LibraryTile.tsx` | `ActiveLoraRow` subscribes to `pendingRefit`, gates `pointerdown` in JS, calls `markPendingRefit(id)` from `pointerup`, applies `.refit-pending` class + `aria-busy` + tooltip on the track. |
| `web/app/globals.css` | `.lora-active-row.refit-pending .lora-strength-track / -thumb / -value` rules — `pointer-events: none`, `cursor: progress`, dimmed thumb + value. |

## Future work (v2)

Server-acked unlock: `backend.py` sends a `lora_refit_done {id, strength}` WS message after each refit, the client clears `pendingRefit` on receipt with the current timeout as a failsafe. That removes the timing approximation but needs protocol coordination and a companion vendor sync — leaving it for a follow-up.

## Verification

1. **Single commit.** Drag a LoRA slider, release. Track dims for ~400 ms, then becomes interactive again. Trying to grab it during the lock does nothing.
2. **Rapid commits.** Drag → release → immediately try to drag again. Second drag is blocked; release the first, wait, then drag again successfully.
3. **MIDI bypass.** Twist a learned MIDI knob — the slider value continues to update visually and the row does NOT enter the locked state. (No `markPendingRefit` from the MIDI path.)
4. **Disable / re-enable a LoRA mid-lock.** Disable while pending → re-enable later → slider is interactive (the timer auto-fires within 400 ms of the original commit either way).

### DIFF
--- a/demos/realtime_motion_graph_web/web/app/globals.css
+++ b/demos/realtime_motion_graph_web/web/app/globals.css
@@ -1880,6 +1880,35 @@ body[data-mode="graph"]::before {
   opacity: 0.7;
 }
 
+/* Refit-pending lock state. After the dispatcher commits a strength
+ * change (DEBOUNCE_MS after the last pointermove), the LoRA's slider
+ * locks for REFIT_LOCK_MS while the server's TRT refit settles —
+ * rapid commits during that window would queue refits the pipeline
+ * can't service in real time, resulting in slider movements that feel
+ * disconnected from audible output. The track-fill keeps animating;
+ * only input is gated.
+ *
+ * Input blocking is done in JS — every pointer entry point
+ * (ActiveLoraRow, DesktopEdgeDrag, useLoraFaderDrag) calls
+ * dispatcher.isLocked(id) on pointerdown and early-returns. The CSS
+ * here is the visual layer only: cursor:progress gives the standard
+ * "system is working" cue, and opacity dims the thumb/value so the
+ * row reads as inactive without obscuring the fill value the operator
+ * is waiting on. Deliberately NOT pointer-events:none on the track —
+ * that would also kill the title-attribute tooltip and cursor:progress
+ * itself on the track element. */
+.lora-active-row.refit-pending .lora-strength-track {
+  cursor: progress;
+}
+.lora-active-row.refit-pending .lora-strength-thumb {
+  opacity: 0.55;
+  transform: none;
+  box-shadow: none;
+}
+.lora-active-row.refit-pending .lora-strength-value {
+  opacity: 0.45;
+}
+
 /* ============================================================================
    Buttons — single shared "outline + accent on hover" language. The Seed /
    Pause / Send buttons inside the drawer all share this look so the whole
--- a/demos/realtime_motion_graph_web/web/components/Performance/DesktopEdgeDrag.tsx
+++ b/demos/realtime_motion_graph_web/web/components/Performance/DesktopEdgeDrag.tsx
@@ -200,6 +200,16 @@ export function DesktopEdgeDrag({ side }: Props) {
       // Defensive: if no LoRA is bound (or a fast LoRA disable race lands
       // `data-empty` mid-press), don't capture — drag is a no-op.
       if (el.dataset.empty === "true") return;
+      // Refit-lock gate. For the LoRA-side edge bars (not the top
+      // tempo bar), refuse a new drag while a prior commit's refit is
+      // still in flight on the bound LoRA. Same chokepoint as the
+      // ActiveLoraRow handler — see dispatcher.isLocked. The top bar
+      // does not route through this dispatcher so it's unaffected.
+      if (!isTop) {
+        const ids = Array.from(useLoraStore.getState().enabled);
+        const id = ids[slotIndex];
+        if (id && loraStrengthDispatcher.isLocked(id)) return;
+      }
       isDragging = true;
       setDragging(true);
       dragInitialValue = readCurrentValue();
--- a/demos/realtime_motion_graph_web/web/components/Performance/LibraryTile.tsx
+++ b/demos/realtime_motion_graph_web/web/components/Performance/LibraryTile.tsx
@@ -406,6 +406,13 @@ function ActiveLoraRow({ entry }: { entry: LoraCatalogEntry }) {
   const fallbackStrength = useLoraStore((s) => s.strengths[id] ?? 0);
   const value = typeof strength === "number" ? strength : fallbackStrength;
   const toggle = useLoraToggle();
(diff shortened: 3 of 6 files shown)


BuffMcBigHuge:
Review verdict: **not merge-ready as a fix for the stated LoRA refit latency problem**.

The patch is mechanically sound and `npm.cmd run typecheck` passes locally, but the behavior does not line up with the PR's own latency contract.

Findings:

1. **[P1] The lock clears before the refit is likely done.**
   `web/engine/lora/dispatcher.ts` already waits `DEBOUNCE_MS = 300` before writing the committed LoRA value into `sliderValues`, which is where `useParamSync` picks it up. This PR starts the 400 ms lock at `pointerup` in `web/components/Performance/LibraryTile.tsx`, so the visible lock can clear roughly 100 ms after the debounced commit reaches the engine. With the stated 270-295 ms server refit, the real busy window after release is closer to 570-600 ms, not 400 ms.

2. **[P1] First-drag values can still stream before the debounce path exists.**
   `useParamSync` falls back to `useLoraStore.strengths` when `perf.sliderValues` does not contain `lora_str_<id>`. The dispatcher updates `useLoraStore.strengths` immediately on every pointer move. Because dynamic LoRA keys are not present in `DEFAULT_SLIDER_VALUES`, the first drag after enable/reset can still send intermediate values at the param-sync cadence before the 300 ms dispatcher commit has seeded `sliderValues`.

3. **[P2] The scope misses other user pointer controls.**
   The pending-refit gate is only installed in `ActiveLoraRow`. Other direct user gesture paths still call `loraStrengthDispatcher.set()` without a pending lock, including `DesktopEdgeDrag.tsx` and `useLoraFaderDrag.ts`. If the intent is "user-direct LoRA strength changes", the lock needs to live in the shared dispatcher/interaction layer or be applied to all pointer LoRA controls.

4. **[P2] The tooltip/cursor affordance is unreliable.**
   The PR places `title` and `cursor: progress` on `.lora-strength-track`, then sets `pointer-events: none` on that same element while pending. That likely prevents the hover target from receiving pointer hit-testing, so the explanatory tooltip and progress cursor may not show when the slider is locked.

Suggested acceptance path:

- Seed/maintain `perf.sliderValues["lora_str_<id>"]` when LoRAs are enabled or reset so the dispatcher debounce is authoritative from the first drag.
- Move refit-pending state into the LoRA dispatcher or a shared pointer-control helper so all direct pointer LoRA controls share the same behavior.
- Start the lock when the debounced commit actually flushes, or clear from a future server `lora_refit_done` ack with a timeout failsafe.
- If this remains timeout-only, account for debounce + refit time, not refit time alone.

gioelecerati:
v2 pushed (`a1844d7`) addressing all four review findings:

1. **Lock from commit, not pointerup.** `markPendingRefit` now fires inside the dispatcher's debounce timer (`engine/lora/dispatcher.ts`), so the 400 ms window opens at the moment the value lands in `sliderValues` and is on its way to the engine — not 300 ms earlier at pointerup, which is what caused the early-clear v1 bug you flagged.

2. **Seed `sliderValues` on enable.** Exported `seedLoraSliderValue(id, strength)` from the dispatcher and call it from `useLoraStore.{setCatalog, enable, toggle, reset}`. `useParamSync`'s `lora.strengths` fallback path is no longer the source for a freshly-enabled LoRA's first drag.

3. **Share the lock across pointer entry points.** Added `loraStrengthDispatcher.isLocked(id)` and check it in `pointerdown` for `ActiveLoraRow`, `DesktopEdgeDrag` (LoRA side bars only — top bar bypasses), and `useLoraFaderDrag`. MIDI / keyboard / MCP paths are discrete commits and stay ungated by design.

4. **Drop `pointer-events: none`.** The JS guard above is now sufficient for input blocking; the CSS class keeps only `cursor: progress` on the track and the thumb opacity dim, so the `title` tooltip and the progress cursor itself both work again.

`npm run typecheck` clean. Ready for another look.

PR #161, opened 2026-05-28
fix(trt): make profile swap atomic w/ session-fatal raise on load failure

leszko:
## Summary

Real fix for the `'NoneType' object has no attribute 'decode'` pipeline_error observed in demon-public-demo admin **Recent pod errors** (pod-c3b46beed068).

Replaces the previously-closed [#160](https://github.com/daydreamlive/DEMON/pull/160), which only relabeled the symptom downstream — that PR didn't reduce the failure rate, just renamed the crash.

## Root cause

`profile_manager._swap()` is **not atomic**:

```python
# acestep/engine/trt/profile_manager.py
self._purge_stale_vae_encode_cache(new_enc)   # line ~317: evict old
...
_get_trt_vae(new_enc, self._device)            # line ~327: load new — can raise
```

When line 327 raises (typical cause: `engine.create_execution_context()` returns None when the 240 s `vae_encode` profile's ~16 GiB workspace doesn't fit), the cache stays empty and the exception unwinds. The runner keeps going.

The next `VAEDecodeAudio.execute` finds an empty `_trt_vae_cache` → falls back to `handler.tiled_decode(...)` → calls `self.vae.decode(...)` on a `None` VAE (production builds `ModelContext` with `skip_vae=True` for TRT-only mode). That's the bare NoneType error users see.

## The fix

Three coordinated changes:

### 1. `profile_manager._swap`: wrap engine loads, retry once, raise typed error

Both decoder and vae_encode loads now run inside try/except. Failures clear `_loaded_dur` / `_loaded_paths` (the prior engines are gone, so the profile manager is now "nothing loaded") and re-raise as a new typed `TRTProfileLoadError`.

For `vae_encode` specifically, one retry runs after `torch.cuda.empty_cache()` — a workspace-alloc OOM is typically just fragmentation from the eviction we just did, and the cache hand-back of reserved-but-unused blocks reliably recovers. Decoder load is one-shot (no observed failures there yet, retry would be unproven and adds no value).

### 2. `vae_nodes._get_trt_vae`: don't cache None contexts

When `engine.create_execution_context()` returns None, raise BEFORE inserting into `_trt_vae_cache`. Without this, the None entry gets cached and any retry from `_swap` short-circuits on the dict lookup and returns the same poisoned entry forever. With this, the retry path actually re-attempts the load.

### 3. `backend.py:apply_swap`: catch `TRTProfileLoadError`, send `swap_failed`

New `except TRTProfileLoadError` mirrors the existing `EngineNotBuiltError` handler. Sends a `swap_failed` JSON to the client:

```
"GPU memory pressure prevented loading the {component} engine for this
source duration. Try a shorter source clip."
```

The user sees an actionable message instead of waiting for the next tick to crash with an opaque `'NoneType'.decode`.

## Why this is a fix, not diagnostics

Before:
- Session keeps running after `_swap` raises
- Next decode tick crashes with `'NoneType' object has no attribute 'decode'`
- User sees an unexplained WS close
- pod is left healthy but the user got no warning

After:
- `_swap` raises `TRTProfileLoadError` at the swap site
- `apply_swap` catches it, sends typed `swap_failed` to the client
- session terminates cleanly with an actionable message
- empty_cache retry recovers from transient fragmentation without any user impact

The failure rate for sessions hitting workspace-alloc fragmentation should drop substantially (retry case), and the sessions that genuinely can't fit get a real error instead of a crash.

## Test plan

Hard to stage locally (workspace-alloc OOM requires a GPU under real VRAM pressure). On a fresh dev pod:

- [ ] Force `_get_trt_vae` to raise on first call for the 240 s vae_encode profile (monkeypatch / inject); confirm `_swap` retries after `empty_cache` and succeeds on second attempt
- [ ] Force it to raise twice; confirm `TRTProfileLoadError` propagates, `_loaded_paths` is cleared, and `apply_swap` sends `swap_failed` to the client
- [ ] Confirm no behavior change on the happy path (matching-profile swap is still a no-op, mismatched-profile swap still loads cleanly)
- [ ] Run a long-source upload that previously triggered the 240 s engine on a card that can't fit it; confirm user sees the friendly swap_failed message rather than the WS dying

## Files

- `acestep/engine/trt/profile_manager.py` — new `TRTProfileLoadError` class + atomic swap with retry
- `acestep/nodes/vae_nodes.py` — don't cache None ctx
- `demos/realtime_motion_graph_web/backend.py` — catch `TRTProfileLoadError` in `apply_swap`

### DIFF
--- a/acestep/engine/trt/profile_manager.py
+++ b/acestep/engine/trt/profile_manager.py
@@ -52,6 +52,43 @@ class _LoRASnapshot:
     strength: float
 
 
+class TRTProfileLoadError(RuntimeError):
+    """Raised by :class:`TRTProfileManager` when a profile swap fails at
+    engine-load time — typically because ``create_execution_context()``
+    returns None under CUDA-OOM workspace pressure (the 240 s
+    ``vae_encode`` profile reserves ~16 GiB), or because
+    ``DiffusionEngine.load_trt_engine`` raises for the same reason.
+
+    Distinct from :class:`acestep.paths.EngineNotBuiltError` (the engine
+    file is missing) and from runtime decode/encode failures (which fire
+    *after* a successful load).
+
+    By the time this is raised the prior engines have already been
+    evicted and the new ones failed to load — the profile manager's
+    ``_loaded_paths`` / ``_loaded_dur`` are cleared to reflect "nothing
+    loaded." Callers should treat the session as unrecoverable and close
+    cleanly (sending a ``swap_failed`` to the client is the established
+    pattern, mirroring how ``EngineNotBuiltError`` is handled in
+    ``demos/realtime_motion_graph_web/backend.py``).
+    """
+
+    def __init__(
+        self,
+        *,
+        component: str,
+        engine_path: str,
+        cause: BaseException,
+    ) -> None:
+        self.component = component
+        self.engine_path = engine_path
+        self.cause = cause
+        super().__init__(
+            f"TRT {component} engine load failed after profile swap eviction: "
+            f"{engine_path} ({type(cause).__name__}: {cause}). "
+            f"GPU may be under memory pressure; session is unrecoverable."
+        )
+
+
 class TRTProfileManager:
     """Owns the live decoder + vae_encode slots and swaps between profiles.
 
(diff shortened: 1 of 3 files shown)


ryanontheinside (approved):
i will rebase and merge

PR #178, opened 2026-05-29
rtmg/web: named-slot prompt deck (kill A/B crossfade)

hthillman:
## Summary

Three coordinated UX changes to the Tags feature in the realtime motion-graph web demo.

**1. Kill the A/B crossfade. Replace with a named-slot deck.**
The old two-textarea + blend-slider UI read as an abstract implementation detail. Operators rarely manually scrubbed the slider mid-performance; the more common gesture was "I want to switch to a different idea." So:
- One \"active prompt\" textarea, bound to the current slot.
- Strip of named slot chips underneath (\"Dubstep\" / \"Daft Punk\" by default, + to add, double-click to rename, × to delete).
- Tap a non-active slot → the new \`lib/promptDeck.ts\` orchestrator loads the target text into whichever engine slot is currently *inactive*, sends the new A/B pair, and tweens \`prompt_blend\` toward it via the existing smoothing system. The lerp is still there — just no longer operator-visible.
- Server protocol completely unchanged. The deck is a client-only construct mapped onto the existing two-slot engine via ping-pong.

**2. Surface the Strength/Structure dependency.**
Inline caption under the textarea: *\"Hits hardest at high Strength + low Structure.\"* The relationship was tribal knowledge that the UI never communicated.

**3. HeroMacros gets a Prompts zone.**
Compact slot-chip strip in the bottom-center bay (mirrors the Styles fader zone next to it) + a \`⋯\` that opens Full Controls already routed to the Styles tab — new \`dd:open-drawer-tab\` event lets the chip address \"open drawer + select tab\" in one move. To make horizontal room without compressing Tools, the Record / Curve / Full / MIDI buttons are now a **2×2 grid** instead of a vertical 4-stack.

## State model

New store fields on \`usePerformanceStore\`:
- \`promptSlots: PromptSlot[]\` — the deck.
- \`currentSlotId: string\` — which slot is logically active.
- \`physicalSlot: 0 | 1\` — which engine slot (A or B) currently holds the active text.

Switching: load target into the inactive engine slot, \`sendPrompt(A, B)\`, set \`prompt_blend\` target to the inactive slot, swap \`physicalSlot\`. The Smooth toggle's \`smoothMs\` controls the transition duration; with smoothing off the switch is instant.

Defaults seed two slots from the existing promptA/promptB defaults so the deck is non-empty from the first session. Saved-session restore continues to write \`promptA/promptB\` directly; the deck rebuilds on next interaction.

## Test plan

- [ ] Open Full Controls → Styles tab. Verify the new deck UI: one big textarea, slot strip, hint line, Send Tags button. The old A/B slider should be gone.
- [ ] Type in the textarea, click Send Tags — engine adopts the new text (verify via \`window.__demonPromptLog = true\`).
- [ ] Click a non-active slot — engine should fade smoothly to the new prompt (with Smooth on; instant with Smooth off).
- [ ] Click \"+\" to add a new slot — drops into rename mode automatically, becomes the active slot.
- [ ] Hover an existing slot — \`×\` button appears. Click it to delete (no-op on the last remaining slot).
- [ ] Double-click a slot label to rename inline. Enter commits, Esc cancels.
- [ ] On the HeroMacros bay (session active, drawer closed): the Prompts zone should show the slot chips. Tapping a chip switches slots the same as the in-tile click. The \`⋯\` button opens Full Controls already on the Styles tab.
- [ ] Tools cluster on the right of the bay is a 2×2 grid (Record/Curve over Full/MIDI). When the drawer opens, it collapses to a horizontal row (same as before).
- [ ] No regressions on existing prompt-blend MIDI bindings or \`schedule_curves\` automation of \`prompt_blend\`.

## Follow-ups (out of scope for this PR)

- Hold-two-fingers / shift-click on a slot to manually live-blend between two slots (the old slider gesture, now hidden by default).
- Per-slot LoRA presets (today the LoRA strengths are global; would be a natural extension once the deck is settled).
- Saved-session integration: roundtripping the entire deck instead of just A/B.


ryanontheinside (changes requested):
The deck UI itself is genuinely great. The named-slot framing, the bay integration, and the Strength/Structure hint are all real wins. 
My one hard blocker is that this removes controllable prompt blending as a capability.

It comes down to the philosophy we talked through yesterday: DEMON is an instrument, and for an instrument the answer is always *more* control, never less. People perform on MIDI controllers, and blending between two prompts on a knob is an obvious, expressive gesture. That's not an edge case, it's an example of the core that makes this an instrument.

The PR's reasoning is basically "users read the slider as an abstract implementation detail." Fair observation, but that's a communication problem, not a feature problem, and with a handful of user interviews behind it I don't think we have the signal to call it. Before removing something hard-won with explicit use cases because people don't immediately get it, I'd want to ask:

- Is there a better way to *present* the gesture so it reads as musical? (e.g. blend lives on the slots: hold/drag between two chips, not a naked A/B slider.)
- Who's actually confused, and how many? Excising a practical, hard-won feature is a high bar: I'd want a large sample and a supermajority before I'd believe removal beats redesign.

And these two behaviors aren't in tension. A clean tap-to-switch deck and controllable blend (drag/hold between slots) can live in the same design: the deck is the simple default, the blend is the depth underneath. The shift-click / two-finger live-blend you already noted as a follow-up is exactly the right seed.

So my ask: keep this PR's UI wins, but treat the simplification as *configurable depth*, not removal. Simplify the default surface; keep the full blend control reachable. If we simplify, it should be opt-in, not destructive.

Requesting changes on that basis, but I'm excited about where this goes. Happy to whiteboard the combined version anytime.

The more knobs we add, the more this becomes an instrument, which is what differentiates this from other AI music tools.

hthillman:
Super valid @ryanontheinside! I’ll do another round with controllable blending restored.

hthillman:
@ryanontheinside 

crossfader is back with MORE CONTROLLABILITY THAN BEFORE. Both endpoints are popover-pickable, so you can use any pair (A→B, A→C, B→C, etc).

PR #197, opened 2026-05-31
feat(rtmg): adaptive gap-filled decode lead, decoupled from VAE span

ryanontheinside:
The windowed-decode playback lead (how far ahead of the live playhead a fresh slice is written) was max((tick+decode) wall-time EMA, half the fixed VAE decode span), capped at half the span. With the ~1s fixed profile the half-span floor pinned the lead near ~0.5s, so a Strength change landed ~0.5s late, and the EMA tracked per-tick compute rather than the interval between writes, so it was blind to the engine's bursty production stalls and placed slices unstably under load.

This decouples the lead from the decode span and sizes it from the actual production cadence:

- Gap-fill: on an active tick where stream.tick() returns no new generation, re-decode the last latent at the advancing playhead and write a slice (decode+write only; latent_history / last_latent / num_gens are left untouched so the feedback delay-tap and generation accounting stay correct). The inter-write interval collapses to ~1 tick across the whole depth x steps grid.
- Adaptive lead: EMA of the (now ~1-tick) inter-write interval * gain + fixed transit margin + a decaying one-shot stall bump. The bump rises predictively when a rebuild-triggering param changes (steps, RCFG mode, LoRA set) and reactively for any gap whose shortfall beyond the slice width would leave a hole; it decays over a tau held above the lead ceiling so the lead can never shrink faster than the playhead advances (decode_start stays monotonic).

Measured live: steady lead ~0.11s at depth=4/steps=8 and ~0.135s at depth=1/steps=16 (both flat, no wobble), zero coverage holes, win_start monotonic, smooth recovery from rebuild stalls.

### DIFF
--- a/acestep/streaming/pipeline_runner.py
+++ b/acestep/streaming/pipeline_runner.py
@@ -333,21 +333,22 @@ class PipelineRunner:
         self._stall_extra_s = 0.0
         self._stall_release_tau_s = 1.8
 
-        # ----- Predictive prewarm on rebuild-triggering param changes -----
-        # A reactive term can't pre-fill the single ~1s stall on the very tick
-        # a rebuild-triggering param first lands (the slow ``tick()`` and the
-        # new value arrive together, and the loop is blocked inside ``tick()``
-        # so no gap-fill runs). So when we detect such a change we raise
-        # ``_stall_extra_s`` to a learned worst-rebuild estimate BEFORE the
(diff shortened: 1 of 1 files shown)


leszko (approved):
Pushed some fixes from Claude and Codex. Other than that, I think it's a good improvement. Works good for me. Merging.

PR #199, opened 2026-06-01
fix(stream): silence latent off-by-one crashes source swap to short/o…

leszko:
…dd-length clips

Swapping the streaming source mid-session to a clip whose latent frame count T satisfies int(T/25.0 * 25) != T crashed the pipeline with:

  The size of tensor a (409) must match the size of tensor b (410)
  at non-singleton dimension 1

in LatentBlend, when hint-strength blending mixes the silence latent with the source's context latent. This hits ~5% of clip lengths (e.g. a 16.4s / 410-frame mic recording -> int(16.4*25) == 409), so arbitrary user recordings and loops trip it.

Root cause: PipelineRunner._rebuild_silence_latent sized the silence latent by round-tripping the source frame count through seconds (T_frames / 25.0) into EmptyLatent, which converted back with T = int(duration * 25). Float truncation drops one frame for many durations, leaving silence at T-1 while the context latent is T.

Fix:
- EmptyLatent accepts an exact integer `frames` (hidden NodeParam); the duration fallback now uses round() instead of truncation.
- _rebuild_silence_latent passes the source's exact frame count.

Verified by reproducing the exact error on the old path and confirming the blend succeeds for every 1..2000-frame source on the new path.

### DIFF
--- a/acestep/nodes/vae_nodes.py
+++ b/acestep/nodes/vae_nodes.py
@@ -708,6 +708,15 @@ class EmptyLatent(BaseNode):
                     description="Duration (s)",
                     min=1.0, max=600.0, step=1.0,
                 ),
+                NodeParam(
+                    name="frames", type="number", default=0,
+                    description=(
+                        "Exact latent frame count (0 = derive from duration). "
+                        "Set by the streaming runner to avoid the "
+                        "duration round-trip dropping a frame."
+                    ),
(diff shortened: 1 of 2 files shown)


gioelecerati (approved):
LGTM @ryanontheinside if you can take a look

ryanontheinside (approved):

PR #201, opened 2026-06-01
feat(rtmg): soften adaptive decode-lead defaults + make them config-t…

ryanontheinside:
…unable

The playback lead (PipelineRunner._decode_advance_s) self-sized within too wide a band: a 0.05s floor with no slack for GPU contention, and a 1.6s ceiling it could balloon to under any co-resident load (screen capture, the WebGPU display, a second process), producing the audible sawtooth/latency spikes. Pin the defaults to a midway profile and expose the bounds in config.json so an install can retune without a code edit.

Defaults:
- lead_floor_s          0.05 -> 0.25
- decode_lead_ceiling_s 1.6  -> 1.35
- stall_release_tau_s   1.8  -> 1.5   (invariant: tau >= ceiling)

Configurable end-to-end via engine.lead_floor_s / lead_ceiling_s /
lead_release_tau_s: config.json -> lib/config.ts -> protocol.ts ->
useStartSession -> SessionConfig -> StreamingSession.run -> PipelineRunner. Read at session start (parity with vae_window / walk_window_s). The runner clamps tau up to the ceiling so an operator misconfig can't break the monotonic-decode invariant.


leszko (approved):
LGTM

PR #202, opened 2026-06-01
fix(rtmg): reserve config tooltip space on mobile

### LINKED ISSUE #192: popup tooltip in full screen obfuscated import/export button
the layout should match the limited view, with a reserved space for the tooltips

nanookclaw:
## Summary

Fixes #192 by giving the full-screen mobile Config controls a reserved tooltip/readout area instead of relying on hover/focus pseudo-tooltips that get cramped in the sheet layout.

The Import and Export buttons now share named tooltip strings with the readout text, so the visible mobile copy stays in sync with the existing desktop tooltip content. The readout is hidden by default and only shown inside `.mobile-sheet-section` at the same responsive breakpoint used for the mobile sheet, preserving the desktop drawer layout.

I also make the two Config buttons fill a simple two-column row in the mobile sheet and suppress their pseudo-tooltip there, avoiding the obfuscated overlay while keeping the button labels and accessible descriptions intact.

Verification:
- `npm ci`
- `npm run typecheck`
- `git diff --check`

### DIFF
--- a/demos/realtime_motion_graph_web/web/app/globals.css
+++ b/demos/realtime_motion_graph_web/web/app/globals.css
@@ -2543,6 +2543,9 @@ body.drawer-open .hud-help-readout {
   gap: 8px;
   align-items: center;
 }
+.operator-config-readout {
+  display: none;
+}
 .install-section-operator .ws-url-input { width: 220px; }
 /* Density toggle anchors the right cluster — its margin-left: auto pushes
    itself plus the trailing pause button to the far right of the row. */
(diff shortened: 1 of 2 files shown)


gioelecerati:
Hey @nanookclaw thanks a lot for opening this PR! Will review it tomorrow :)

ryanontheinside (approved):
Thanks for this, and welcome! This is the first contribution we've had from outside the core team, so it's genuinely exciting to see, and it's a clean, well-scoped one to start with.

The fix is correct for the right reason. There's already a global @media (pointer: coarse) rule that suppresses the pseudo-tooltip on touch devices, so the bug only surfaces on a fine-pointer + small-viewport "full screen" layout where that rule doesn't fire. Scoping the
new suppression to a size-based media query covers exactly that gap.

PR #204, opened 2026-06-01
Small Fixes

BuffMcBigHuge:
Fix to vae spam logging, fix to modal handling of uploaded track.

### DIFF
--- a/acestep/nodes/vae_nodes.py
+++ b/acestep/nodes/vae_nodes.py
@@ -3,6 +3,9 @@
 from __future__ import annotations
 
 import os
+import sys
+import threading
+import time
 from pathlib import Path
 from typing import Any, ClassVar, Optional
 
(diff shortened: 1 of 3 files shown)


ryanontheinside (changes requested):
This needs a small fix before merge.\n\nThe new inish() path in AlmostReadyDialog calls onContinue(...) and then immediately calls onClose(). All current callers wire onClose to setPending(null), while commitUploadedTrack still intentionally keeps pending until upload success so failed encodes/uploads can be retried. With this PR, pressing Start dismisses the dialog and clears the trimmed upload buffer before the async upload has succeeded; if the upload fails, the user only gets a status message and cannot retry the same trimmed selection.\n\nPlease close the dialog only after commitUploadedTrack succeeds, or otherwise preserve/restore retry state on failure.

BuffMcBigHuge:
Thanks @ryanontheinside — good catch. Fixed in `AlmostReadyDialog.finish()`.

The dialog no longer calls `onClose()` itself. It now `await`s `onContinue(...)` and lets the parent own dismissal: `commitUploadedTrack` already clears `pending` upstream on success (which unmounts the dialog) and intentionally keeps `pending` on a failed encode/upload. By dropping the eager `onClose()`, a failed upload now leaves the trimmed selection intact and retryable instead of being thrown away with only a status message.

The double-submit guard (`submittedRef`) is preserved to block re-entry while the commit is in flight, and it's reset in a `finally` so a failure that keeps the dialog mounted can be retried with one more press of Start. `onContinue`'s type was widened to `void | Promise<void>` to support the await.

ryanontheinside (approved):

ryanontheinside (changes requested):
#249 fixed the vae logging more directly and was merged. the other changes here are likely still valuable but should be isolated

PR #205, opened 2026-06-01
feat(rtmg): midi mapping

ryanontheinside:
Reworks MIDI mapping for the RTMG web demo.

## What's new
- **Per-binding CC controller type:** `absolute` (knob/fader position = value), `relative` (endless encoder), `centered` (joystick/pitch spring), `unipolar` (mod-wheel spring).
- **Pitch-bend (0xE0) support** — joysticks and pitch wheels are now mappable.
- **Enum targets** (DCW mode, wavelet, RCFG, pipeline depth, key, time signature): map a knob (quantized sweep) or a pad (cycle on press).
- **One shared right-click MIDI menu** for every control — mapping and mode both live there.

## ⚠️ Behavior change — CC default mode
- **Before:** a mapped CC was delta-tracked — the knob nudged the slider from its current position, so there was no takeover snap (but absolute position was never honored 1:1).
- **Now:** a CC with no stored mode defaults to **`absolute`** — the knob's physical position sets the value directly.

Maps saved before this PR have no `ccMode`, so they will **snap to the knob position on the first twist**. Right-click the control → pick a mode (e.g. `relative` for endless encoders) to change it. The old delta-tracked-absolute mode no longer exists.


leszko (approved):
Pushed a few minor fixed. Test, it all works good. Merging.

PR #209, opened 2026-06-02
feat(rtmg/web): host-supplied action gate for track-change + upload

hthillman:
## Summary

Adds `useActionGate()` — a context-supplied async predicate that hosts can use to intercept the two highest-intent user actions in the Performance UI (track switch + audio upload) *before* the underlying store mutation runs. Default returns true, so standalone DEMON demos and any consumer that doesn't mount a provider are unaffected.

The catalyst is daydream-public-demo's signup-conversion problem: anonymous users currently switch tracks and upload audio with zero friction, and we see a 1–2% page→signup conversion rate. A hard gate at those two surfaces should lift conversion meaningfully — but the click handlers live here, so DEMON has to expose the seam. The consumer side (a SignUpGateDialog + an `<ActionGateProvider value={...}/>` that returns false for anonymous users and pops the dialog) ships in the next demon-public-demo PR, gated on a sync-ui bump from this one.

## What's gated

- **AudioSourceCrate** — desktop crate: top-dock Upload button, the per-track sleeve in the fan, the upload sleeve at the bottom of the fan.
- **LiteTrackCarousel** — mobile native-select: track-pick branch and the `__upload__` sentinel. Mic recording is NOT gated; only track-pick and upload are.
- **TrackPicker** — CORE panel inline picker: `RefSelect.onSelect` and `onUpload`.

The "pick another" re-trigger inside `AlmostReadyDialog` is intentionally not gated — by the time a user sees that dialog they've already cleared the first gate.

## Why a context and not a prop

These click sites sit deep in three components and any future surface that adds a new gated affordance. Threading a prop through every parent adds maintenance cost the host shouldn't have to pay. A single provider at the Performance boundary covers them all.

## Test plan

- [ ] Standalone DEMON demo at `demos/realtime_motion_graph_web/web/` still works — no provider mounted, default gate returns true, every track-change + upload runs exactly as before
- [ ] Type-check passes (`tsc --noEmit`)
- [ ] Consumer side in demon-public-demo gates correctly once it lands

### DIFF
--- a/demos/realtime_motion_graph_web/web/components/Performance/AudioSourceCrate.tsx
+++ b/demos/realtime_motion_graph_web/web/components/Performance/AudioSourceCrate.tsx
@@ -9,6 +9,7 @@ import {
   type DecodedFixture,
   type StemSourceMode,
 } from "@/engine/audio/loadFixture";
+import { useActionGate } from "@/hooks/useActionGate";
 import { useSeedUserUploads } from "@/hooks/useSeedUserUploads";
 import { commitUploadedTrack } from "@/lib/audio/commitUploadedTrack";
 import { useUploadOnboardingHint } from "@/hooks/useUploadOnboardingHint";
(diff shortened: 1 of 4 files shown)


gioelecerati (approved):
LGTM - could be worth it add the microphone input to the gate

hthillman:
Good call — done in d273be3. `ActionGateKind` now includes `"mic"` and both mic entry points (top-dock Mic button + mobile select `__mic__` sentinel) are gated. The consumer-side update — copy + event types — also went out on demon-public-demo#303 just now.

PR #219, opened 2026-06-03
feat: add websocket startup telemetry

j0sh:
Add browser-side WebSocket lifecycle tracing and an optional init_ack from the backend so client startup failures can be correlated with pod logs.

The client now records connect/open/config/ack/ready/error/close timing per attempt, stores orphan pre-ready traces in session state, and opts into telemetry with telemetry_version=1. The backend replies with init_ack only for opted-in clients after it binds session_id/client_id context, preserving old-client compatibility.

### DIFF
--- a/demos/realtime_motion_graph_web/web/engine/protocol.ts
+++ b/demos/realtime_motion_graph_web/web/engine/protocol.ts
@@ -21,6 +21,7 @@ import {
   SLICE_FLAG_DELTA,
   SLICE_HDR_SIZE,
   type AudioSlice,
+  type InitAckMessage,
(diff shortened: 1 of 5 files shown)


ryanontheinside (approved):
One suggestion: add a short comment on `telemetry_version` in `types/protocol.ts` noting that nothing leaves the browser in this repo (the trace is local; downstream deployments may consume it). Verified the diff has zero egress, but a grep-stopper there saves OSS users the archaeology.

j0sh:
Added grep-stopper comment in 60aac013c54a6fb8a5069cfee87ebac8d9e13d5d @ryanontheinside

PR #240, opened 2026-06-10
Ryanontheinside/feat/latency/06 near playhead repatch

ryanontheinside:
# Knob-to-ear latency stack — plan + progress

Untracked working file. Source review: 2026-06-09 (memory: knob-to-ear-code-levers).
Base: main @ dc06b19. Stack is linear; each branch is based on the previous.
Gates per branch: `.venv/Scripts/python.exe -m pytest tests/unit` + typecheck of touched files.
Golden harness: ONLY at the end, on the stack tip, with Ryan's explicit OK.
No pushes / PRs without Ryan's explicit per-action approval.
Out of scope: shadow migration (lives in DEMON_alt2 / exp branches, own landing plan).

## Stack (in order)

1. `ryanontheinside/perf/latency/01-cuda-event-timing` — [DONE: pending]
   diffusion_backend.py:102-119 two full-device cuda.synchronize per tick exist only
   to time last_tick_ms; ace_backend.py render_window has a third before `.cpu()`.
   Replace with CUDA events (lazy elapsed_time read; one-tick-stale timing is fine —
   verify nothing reads last_tick_ms same-tick for logic).

2. `ryanontheinside/perf/latency/02-shared-curve-cache` — [DONE: pending]
   stream.py set_shared_curve stores CPU fp32; _eff_shared re-normalizes every read;
   _resolve_slot_curves .to(device) per slot per step. Pre-normalize + device-cast at
   the setter (precedent: set_channel_gain_tensor), lazy-cast if device unknown at set
   time. Slot-field fallback path keeps read-time normalize.

3. `ryanontheinside/perf/latency/03-x0-gate-no-readback` — [DONE: pending]
   stream.py:1226-1231 eff_strength.abs().any().item() = host fence per slot per step.
   Compute nonzero-flag once at set_shared_curve / slot init; read the cached bool.

4. `ryanontheinside/perf/latency/04-batched-small-h2d` — [DONE: pending]
   stream.py:721-723 per-element timestep writes -> one pinned staging copy_.
   model_adapter.py:166-173 per-forward torch.tensor(timesteps) + torch.ones attn
   mask -> cached/reused buffers.

5. `ryanontheinside/perf/latency/05-prompt-encode-unlocked` — [DONE: pending]
   streaming/session.py set_prompt (1396), timbre/structure apply (1212, 1660),
   prompt_blend swap (966-990): GPU encode runs under state._lock while the runner
   takes the same lock every tick (ace_backend read_knobs:541, has_pending_refit:566).
   Encode outside the lock, swap under it; generation counter so a stale encode never
   overwrites a newer one.

6. `ryanontheinside/feat/latency/06-near-playhead-repatch` — [DONE: pending]
   pipeline_runner.py: after a fresh produce, additionally render+patch a window at
   playhead + transit margin (not just playhead + lead, floor 0.25s @ :239). Latent
   state covers all positions (gap-fill proves it). Kills the lead floor as the
   audibility floor for knob changes. Mind the monotonic-decode comment (:245-248) —
   repatch is a deliberate, separate write path. Default behavior decided here;
   measure with golden latency report.

## Parity traps
- NEVER remove/reorder randn_like calls (RNG draw order parity; see memory).
- Curve pre-cast must produce byte-identical math to read-time .to(dtype) casts.
- ace_backend timing fields (last_tick_ms/last_dec_ms) feed the runner trace + golden
  latency report; keep field semantics.

## Status log
- 2026-06-09: plan written, stack not started.
- 2026-06-09: all six branches committed (01:70ebbff 02:e13f03f 03:8b08147
  04:0d225e6 05:0cddfae 06:55f06c5), each green on tests/unit (149 passed).
  Commit messages must go via `git commit -F <file>` — the settings deny
  list pattern-matches message text (e.g. "*rd *", "*nc *") and multiline
  commands choke the harness.
- Next: GPU end-to-end on stack tip (drain parity script, live smoke,
  golden harness last — Ryan approved the golden run 2026-06-09).
- 2026-06-09 e2e: drain parity PARITY OK (8 latents bit-identical, ODE+SDE);
  full golden 12/12 PASSED at tip in 4:51. knob_step initially showed
  first=218/full=578 (vs 234/594 baseline) — repatch WAS firing (109
  lat_nearpatch lines, close~0.12-0.18s) but _action_audible took the
  first-ARRIVING qualifying slice; repatch slices arrive later with
  earlier starts. Fixed metric to min-over-slices of max(arrival,
  playhead-reaches-start) (0539d71); trace line amended into 3ba48f6
  (06 tip moved 55f06c5 -> 0539d71). audible_full is structurally
  playhead-to-frontier distance; the lever for it is shadow migration,
  not this stack. Re-validation run + final full golden pending.
- 2026-06-09 FINAL: corrected metric run: knob_step first=58ms (was 234
  baseline / 218 old-metric), full=578 (structural). prompt_change:
  next_slice=31ms during the encode (ticks never stall, branch 5 works);
  prompt ack 156->313ms (encode now shares GPU with ticks — right trade).
  Final gates at tip 0539d71: tests/unit 149 passed; golden 12/12 in 4:48.
  STACK COMPLETE, local only — pushes/PRs need Ryan's per-action approval.

### DIFF
--- a/acestep/engine/model_adapter.py
+++ b/acestep/engine/model_adapter.py
@@ -104,6 +104,14 @@ class ACEAdapter:
 
     def __init__(self, pipeline):
         self._pipeline = pipeline
+        # Per-shape reuse caches for the eager (non-TRT) forward:
+        # (pinned-host, device) timestep buffer pairs keyed by B, and
+        # the all-ones attention mask keyed by (B, T). Both were
+        # allocated fresh on every forward. Shapes only vary with the
+        # ring-buffer fill level and the source duration, so the dicts
+        # stay tiny; cleared wholesale if they ever grow past 16.
+        self._t_bufs: dict = {}
+        self._attn_ones: dict = {}
 
     def build_schedule(self, config, denoise: float, device, dtype) -> torch.Tensor:
         from .diffusion import DiffusionConfig
@@ -163,14 +171,32 @@ class ACEAdapter:
                 ctx_batch=ctx_b,
             )
 
-        t_b = torch.tensor(
-            timestep_list, device=p._device, dtype=p._dtype,
-        )
+        B = xt_batch.shape[0]
+        tb = self._t_bufs.get(B)
+        if tb is None:
+            if len(self._t_bufs) > 16:
+                self._t_bufs.clear()
+            pin = p._device is not None and p._device.type == "cuda"
+            tb = (
+                torch.empty(B, dtype=p._dtype, pin_memory=pin),
+                torch.empty(B, dtype=p._dtype, device=p._device),
+            )
+            self._t_bufs[B] = tb
+        t_host, t_b = tb
+        for i, t in enumerate(timestep_list):
+            t_host[i] = t
+        t_b.copy_(t_host, non_blocking=True)
+
         mask_b = torch.cat(mask_list, dim=0)
-        attn_b = torch.ones(
-            xt_batch.shape[0], xt_batch.shape[1],
-            device=p._device, dtype=p._dtype,
-        )
+        key = (B, xt_batch.shape[1])
+        attn_b = self._attn_ones.get(key)
+        if attn_b is None:
+            if len(self._attn_ones) > 16:
+                self._attn_ones.clear()
+            attn_b = torch.ones(
+                key[0], key[1], device=p._device, dtype=p._dtype,
+            )
+            self._attn_ones[key] = attn_b
 
         out = p.decoder(
             hidden_states=xt_batch,
--- a/acestep/engine/stream.py
+++ b/acestep/engine/stream.py
@@ -736,9 +736,14 @@ class StreamPipeline:
         else:
             bufs["hidden_states"].copy_(xt_io)
 
-        # timestep: one scalar per row.
+        # timestep: stage all rows in the pinned host buffer, then one
+        # async H2D copy — instead of one tiny device write per row.
+        # Ordering with the TRT exec below is the same as the other
+        # input copies (legacy default stream → blocking TRT stream).
+        t_host = bufs["_timestep_host"]
         for i, t in enumerate(timestep_list):
-            bufs["timestep"][i] = t
+            t_host[i] = t
+        bufs["timestep"].copy_(t_host, non_blocking=True)
 
         # encoder_hidden_states: already padded to max_L + catted by
         # the caller. The engine has no ``encoder_attention_mask``
@@ -906,6 +911,11 @@ class StreamPipeline:
         bufs["_eff_T"] = eff_T
         bufs["_T"] = T
         bufs["_out_buf"] = out_buf
+        # Pinned host staging for the per-row timestep scalars (see
+        # _trt_forward). Underscore-prefixed so the bind loops skip it.
+        bufs["_timestep_host"] = torch.empty(
+            B, dtype=bufs["timestep"].dtype, pin_memory=True,
+        )
         self._trt_bufs_cache[key] = bufs
         while len(self._trt_bufs_cache) > self._trt_bufs_cache_max:
             self._trt_bufs_cache.popitem(last=False)
--- a/acestep/engine/stream.py
+++ b/acestep/engine/stream.py
@@ -225,6 +225,12 @@ class _Slot:
     # is normalized once per slot instead of allocating a fresh
     # ``[1, 1, 1]`` tensor on every read (per slot, per step).
     curve_cache: Dict[str, torch.Tensor] = field(default_factory=dict)
+    # Memo for "does this slot field have any nonzero value" — the
+    # x0_target gate used to answer that with a per-slot-per-step
+    # ``.abs().any().item()`` device readback; the flag is computed at
+    # most once per slot here (and without any tensor op for the
+    # common Python-scalar fields).
+    nonzero_cache: Dict[str, bool] = field(default_factory=dict)
 
 
 class StreamPipeline:
@@ -332,6 +338,12 @@ class StreamPipeline:
         # at the setter so callers can pass scalars without thinking
         # about shape.
         self._shared_curves: dict[str, torch.Tensor] = {}
+        # Sidecar of ``_shared_curves``: "any nonzero value" per name,
+        # computed at the setter (no tensor op for scalar sets) so the
+        # x0_target gate in the step loop never has to read a tensor
+        # back from the device. Maintained strictly in lockstep with
+        # ``_shared_curves`` — see set_shared_curve / _curve_nonzero.
+        self._shared_curve_nonzero: dict[str, bool] = {}
 
         # Channel guidance: a ``[1, T, 64]`` per-channel gain applied to
         # ``xt`` before each forward pass. Lives in its own field rather
@@ -1224,16 +1236,11 @@ class StreamPipeline:
 
             # ``x0_target_strength`` path: blend toward a target latent
             # at scalar (or per-frame curve) strength, gated to the
-            # refinement half.  Preserving the historical "strength==0
-            # falls through to the fast path" behavior — checks the
-            # effective (shared override or slot field) strength via a
-            # tensor.any() sync, which costs one host-device fence per
-            # slot per step but lets the gate stay tensor-safe.
-            eff_strength = self._eff_shared(slot, "x0_target_strength")
-            strength_active = (
-                eff_strength is not None
-                and bool(eff_strength.abs().any().item())
-            )
+            # refinement half. Preserves the historical "strength==0
+            # falls through to the fast path" behavior; the nonzero
+            # check reads cached flags (maintained at set time), not
+            # tensor data, so the gate costs no host-device fence.
+            strength_active = self._curve_nonzero(slot, "x0_target_strength")
             scalar_x0_target = (
                 req.x0_target is not None
                 and strength_active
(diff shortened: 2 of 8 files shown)


BuffMcBigHuge (commented):
# Full review — read against #238 (`marco/feat/latency`)

Both PRs came out of independent same-week latency reviews of the same stack, so I diffed both branches side by side against the common base `dc06b19`. Summary: roughly half of each PR is the same three levers implemented twice; the other halves are disjoint and compose. Net of a merge, almost everything in both survives — details and a suggested reconciliation below.

## Where we built the same thing twice

**1. `x0_target_strength` gate readback** (8b08147 vs #238 §4.4)
Same diagnosis (per-slot-per-step `.abs().any().item()` fence in the integration loop), same fix (set-time flags). Yours is the better cut: `_curve_nonzero` is generalized over curve names, the slot-field flag memoizes lazily and costs no tensor op for Python-scalar fields; ours special-cases x0 with an eager slot-init flag. **Take yours; ours drops.**

**2. Shared-curve hot path** (e13f03f vs #238 §4.4)
Same target, different halves of the fix:

- Yours: device-cast at the setter, plus `slot.curve_cache` for the slot-field fallback path (our version still re-normalizes that path on every read).
- Ours: setter dedup — re-setting the same scalar / tensor object is a no-op, which matters because `ace_backend` pushes the full knob state every tick, so without dedup the setter still re-normalizes once per tick — plus a read-side device+dtype cast cache, and an `ace_backend` SDE-curve rebuild key `(amplitude, periodicity, src_T)` that makes the dedup actually fire for the tensor curve.
- Merged shape: your setter device-cast + `curve_cache`, our setter dedup + SDE-curve key. On dtype: you leave the per-step `.to(dtype)` at the consumer boundary (still an alloc per step when dtypes differ); we cached that cast too — worth keeping exactly one mechanism for it in the merge.

**3. Timestep H2D in `_trt_forward`** (0d225e6 vs #238 §4.5)
Yours is strictly better: pinned staging buffer cached in `bufs` + `non_blocking` copy, vs our fresh pageable `torch.tensor` per forward. (Our own writeup listed pinned+async as further-proposal #5 — you built it.) You also covered the eager path in `model_adapter`, which we didn't touch. **Take yours.** One invariant worth recording: reusing the pinned host buffer across forwards is safe because the per-tick emission D2H drains the stream before the next tick writes it again — if emission ever stops being per-tick, this needs double-buffering.

**4. Host syncs / timing — two non-overlapping halves of the same problem.**
70ebbff replaces the *measurement* syncs (the two per-tick full-device `torch.cuda.synchronize()` in `diffusion_backend` + the decode bracket in `ace_backend`) with CUDA events, but keeps the TRT enqueue syncs. #238 §4.3 removes the *enqueue* syncs (`_trt_stream.synchronize()` in `_trt_forward` and `_trt_vae_decode`, same blocking-stream ordering argument your commit message makes) but keeps the measurement syncs. Each PR removed exactly the syncs the other kept. Merged, the tick has **no** host sync left except the natural emission `.cpu()` — and your events fix the caveat we documented (with #238 alone, `last_tick_ms` degrades to enqueue time; with your events it's true GPU time again, which matters because the lead controller and golden trace consume it). Your lazy event resolution assumes the render's D2H already drained the stream — that still holds under #238's sync removals, so the composition is sound.

## What only #240 has — the half we didn't build

- **Near-playhead repatch (3ba48f6)** is the headline. Our review called the playback lead "the dominant audible term" but only proposed lead-floor tuning (with the underrun risk that implies); the repatch is the better idea — keep the floored frontier write for stall safety, opportunistically rewrite close-in, let a late landing lose harmlessly. And the two PRs compose multiplicatively here: #238 shortens knob→fresh-latent (queue_cap=1: −5 ticks measured on per-slot params at d4/s8; same-tick delivery: −1 tick on every generation), this PR shortens fresh-latent→ear (~160 ms of buffer distance per your 218→58 ms numbers). On a merged tip, an ODE denoise change gets both.
- **Encodes off `state._lock` (0cddfae)**: our further-proposal #8 territory, done properly. The snapshot/encode/commit + `cond_epoch` retry design reads correctly to me. Two small notes: (a) `set_prompt` writes `state.time_signature` during the snapshot phase, so a 5×-lost prompt still leaves the ts override applied — probably fine, just asymmetric with the rest of the commit; (b) the 5-strikes drop logs and skips `PromptApplied`, so clients waiting on the ack must tolerate a missing event (rare in practice, worth knowing).
- **Golden metric fix (0539d71)**: necessary the moment slice starts stop being arrival-monotonic, and min-over-`max(arrival, reach)` can't be gamed by late arrivals. Good catch — the e2e harness from #238 has the same first-qualifying-slice assumption in its knob→fresh-generation metric and will need the same fix.
- Eager-path buffer reuse in `model_adapter` (pinned timestep pair + cached attn-ones).

## What only #238 has

queue_cap=1 (the single biggest engine-side win), same-tick latent delivery, the TRT enqueue sync removals, the steering zero-fill dirty flag, the schedule-cache LRU bound, the SDE-curve rebuild cache, and the `tests/e2e_latency/` harness (engine-level knob→latent change detection + headless full-stack timing) — which complements your golden-runner work as the build-to-build diffing tool.

## Merge mechanics

`git merge-tree` of the two heads: **only `acestep/engine/stream.py` conflicts** — everything else (including `ace_backend.py`, which both touch) auto-merges. The conflicting hunks are exactly the three duplicated levers above, and the resolution writes itself given the picks: your `_curve_nonzero` + pinned timestep staging + `curve_cache`, our setter dedup + queue_cap + same-tick delivery + TRT sync removals + LRU + steering flag.

Suggested order: #238 already has an approval — land it, rebase this stack on it. 01 (events), 05 (session), 06 (repatch), 07 (golden) rebase clean since #238 doesn't touch those files; 02/03/04 shrink to the deltas above. Happy to do the `stream.py` reconciliation from either side.

One caveat on numbers: both PRs measured against base independently — your 58 ms `audible_first` doesn't include #238's queue-cap/same-tick wins, our 172 ms knob→fresh-generation doesn't include your repatch. The merged tip deserves a fresh golden + e2e run; the bit-exactness probes from both PRs should still pass, since neither side reorders RNG draws.

Great work — genuinely complementary stacks.

BuffMcBigHuge:
@ryanontheinside — full comparison of this stack against [#238](https://github.com/daydreamlive/DEMON/pull/238) (`marco/feat/latency`). I diffed both branches against `dc06b19` and ran a test merge.

## TL;DR

Same week, same stack, two independent reviews. Roughly **half overlaps** (same diagnosis, different implementations); **half is disjoint and composes cleanly**. A merged tip keeps almost everything from both PRs. Only **`acestep/engine/stream.py`** conflicts — everything else auto-merges.

---

## Duplicated (pick one implementation per lever)

| Lever | #240 | #238 | Recommendation |
|---|---|---|---|
| **x0 gate readback** | `_curve_nonzero` — generalized, lazy, no tensor op for scalar fields | x0-only eager slot flag | **Take #240** |
| **Shared-curve hot path** | device-cast at setter + `slot.curve_cache` | setter dedup (same scalar/tensor object = no-op) + SDE rebuild key in `ace_backend` + read-side dtype cache | **Merge both halves** — #240 fixes slot-field re-normalize; #238 fixes per-tick re-push when knobs are steady |
| **Timestep H2D** | pinned staging in `bufs` + `non_blocking`; eager path in `model_adapter` | pageable single `torch.tensor` copy | **Take #240** (our writeup listed pinned+async as a future proposal — you built it) |
| **Host syncs** | removes *measurement* syncs → CUDA events | removes *enqueue* syncs in TRT forward/VAE | **Both** — each PR removed exactly what the other kept; merged = zero host sync in the tick loop, and your events restore true GPU timing for the lead controller |

---

## Only in #240 (we didn't build these)

- **Near-playhead repatch** — the better answer to the playback-lead audibility floor we identified but only proposed floor-tuning for. Frontier write keeps stall safety; close write is opportunistic. Composes with our engine wins.
- **Prompt/timbre encodes off `state._lock`** — snapshot/encode/commit + `cond_epoch` retry. Our further-proposal #8 territory, done properly.
- **Golden `audible_first` fix** — necessary once slice starts stop being arrival-monotonic; min-over-`max(arrival, reach)` is the right metric.
- Eager-path buffer reuse in `model_adapter`.

Your numbers (218→58 ms `audible_first`) were measured without our queue-cap / same-tick wins — merged tip deserves a fresh golden run.

---

## Only in #238 (you didn't build these)

- **`queue_cap=1`** — biggest per-slot engine win (−5 ticks measured at d4/s8)
- **Same-tick latent delivery** — −1 tick on every generation
- **TRT enqueue sync removals** (`_trt_forward`, `_trt_vae_decode`)
- Steering zero-fill dirty flag, schedule-cache LRU bound
- **`tests/e2e_latency/`** harness — engine-level + headless full-stack timing for build-to-build diffing (complements your golden work; note: our first-qualifying-slice metric will need your golden fix once repatch lands)

---

## Suggested landing

1. Land #238 first (already approved).
2. Rebase this stack on it — commits 01/05/06/07 should rebase clean; 02/03/04 shrink to the deltas above.
3. Hand-resolve `stream.py` using the picks in the table.
4. Fresh golden + e2e on the merged tip.

Happy to do the `stream.py` reconciliation from either side. Bit-exactness probes from both PRs should still pass — neither side reorders RNG draws.

PR #251, opened 2026-06-11
Added mintlify docs.

BuffMcBigHuge:


ryanontheinside:
Approved with changes. I applied a few drift fixes after the setup-experience work landed on main, plus a small docs-role cleanup for the Mintlify app. This should be mergeable once the deployment plan for the Mintlify docs is established.

BuffMcBigHuge:
@ryanontheinside I think adding some audio samples would be good as well.

PR #256, opened 2026-06-12
fix(streaming): stop raw-source audio bleed on slow/congested links

leszko:
## Problem

Listeners intermittently hear the **raw input** audio instead of the processed stream — worst on remote pods. The client buffer always holds the source track; denoised slices patch over it just ahead of the playhead, so any slice landing *behind* the playhead bleeds raw source. Three independent causes:

1. **Stale playhead anchors** — under congestion, queued `playback_pos` reports re-anchor the server's playhead clock ever further into the past.
2. **Client slice-apply backlog** — per-slice main-thread cost (listener fan-out + LUFS DSP) makes a busy/throttled tab apply slower than the server produces, so the receive queue grows unbounded.
3. **Bandwidth deficit** — the ~1–2.5 MB/s slice stream buffers tens of seconds on a slow link (SSH/IDE tunnel) where the server can't see it.

## Fix

- **Staleness compensation** — client stamps `params.client_time`; server estimates per-report queueing delay and projects the playhead anchor forward. Recv loop coalesces backlogged params (newest wins); client skips ticks when `ws.bufferedAmount` backs up.
- **Landing-lead feedback** — client reports worst slice landing lead (`params.slice_lead_s`); server raises its transport lead to cover deficits, decaying only while reports show headroom.
- **Flow control** — client acks received slice bytes (`params.slice_bytes_rx`); server skips slice emission pre-encode while in-flight exceeds `DEMON_SLICE_WINDOW_BYTES` (default 256 KiB), with a 2 s queue-age backstop. Old clients keep legacy behavior.
- **Cheaper slice apply** — mirror-change notifications trailing-throttled to 10 Hz; LUFS chunk map maintained only while the matcher is enabled.

Three new optional `params` fields (registry + regenerated TS types).

## Validation

`scripts/flow_control_harness.py` (new): bufferbloat proxy + headless client measuring landing leads end-to-end. Direct / 500 KB/s / 200 KB/s links all hold positive leads after the controller settles. `pytest tests/unit`: 272 passed (9 new); `npm run typecheck` clean.


ryanontheinside (commented):
I rebased this branch on main locally and saw the updated PR head after your force-push. I used the onboard headless MCP for a baseline run, then added focused repros for the new feedback paths below.

demos/realtime_motion_graph_web/ws_adapter.py
```diff
@@ -1525,31 +1667,62 @@ def _recv_binary_payload(fail_type: str):
...
-                            _dispatch_message(data, ws.recv, "ws")
-                        except Exception as exc:
-                            logger.exception(
-                                "ws_dispatch_error error={}", exc,
-                            )
+                        if (
+                            isinstance(data, dict)
+                            and data.get("type") == "params"
+                        ):
+                            pending_params = data  # newest wins
```
ryanontheinside, line 1738: This coalesces `params` by replacing the whole pending message with the newest one, which is fine for `playback_pos`, `client_time`, and cumulative `slice_bytes_rx`, but it drops the semantics of `slice_lead_s`. The wire contract says `slice_lead_s` is the worst observed lead since the previous params report, and the server widens lead based on that worst value.

I reproduced this with a queued sequence of params whose leads were `-1.25`, omitted, then `0.40`: this code dispatches `0.40`, but the feedback controller needs to see `-1.25`. Please preserve the latest snapshot while carrying forward the minimum finite `slice_lead_s` across coalesced params.

demos/realtime_motion_graph_web/web/hooks/useParamSync.ts
```diff
@@ -81,7 +82,11 @@ export function useParamSync() {
       // Playback position is *seconds* (raw audio.positionSec), not a 0..1
       // ratio. The server uses absolute time for curve sampling.
       const playbackSec = session.player.positionSec;
-      session.remote.sendParams(raw, playbackSec);
+      // Worst slice landing lead since the previous tick (null when no
+      // slice arrived). The server widens its playback lead until these
+      // stay positive — see sliceLeadTracker.ts.
+      const sliceLead = takeSliceLead();
```
ryanontheinside, line 88: `takeSliceLead()` clears the worst lead before we know whether `sendParams()` actually writes to the socket. In the new backpressure path, `sendParams()` can return early when `ws.bufferedAmount > PARAMS_BACKPRESSURE_BYTES`, so a negative lead sample is discarded locally and never reaches the server.

I reproduced this by setting the tracker to `-0.75` and using a fake websocket with `bufferedAmount = 9000`: no params payload was sent, and the local worst lead was reset to `null`. The fix should make `sendParams` return a success boolean, or otherwise only clear/rearm this tracker after a successful `ws.send`.

packages/demon-client/audio/AudioPlayer.ts
```diff
@@ -862,8 +890,19 @@ export class AudioPlayer {
...
+    // throttled fire below; it doesn't need a counter.
+    this._scheduleMirrorNotify();
+  }
+
+  /** Trailing-throttled mirror-change notification (~10 Hz). See the
+   *  _notifyTimer comment for why per-patch synchronous fires are not
+   *  affordable on the slice path. */
+  private _scheduleMirrorNotify(): void {
+    if (this._notifyTimer !== null) return;
+    this._notifyTimer = setTimeout(() => {
```
ryanontheinside, line 902: This behavior change currently breaks the existing web unit suite. After this throttling change, `npm run test:unit -- audioPlayer.test.ts` fails in `AudioPlayer > notifies mirror listeners on writes and swaps`: the test expected 3 listener fires after `patch`, `addDelta`, and `swap`, but got 1 because patch/addDelta are now delayed/coalesced.

If the throttling is intentional, please update/add tests with fake timers to lock in the new contract. Right now the PR leaves a failing test in the repo.

demos/realtime_motion_graph_web/ws_adapter.py
leszko: Addressed in 45bcce29d0780396f94393100ace439347802cb1

demos/realtime_motion_graph_web/web/hooks/useParamSync.ts
leszko: Addressed in 0a3d9e50b52bf70378d2f40ad7bc21b8407ad30e

packages/demon-client/audio/AudioPlayer.ts
leszko: Fixed in 7cf64744ec9cb2c7387b728cbfbd775d173024e2

ryanontheinside (approved):

PR #267, opened 2026-06-16
refactor(client/config): portable config.json via @demon/client

ryanontheinside:
## What

Plan 01 of the client-SDK refactor: move the portable `config.json` operator-defaults (schema, `DEFAULT_CONFIG`, pure transforms, loader, and the `RtmgConfig` → wire `SessionConfig` mapping) out of the web app into `@demon/client` (`packages/demon-client/config/`), so web / M4L / VST share one implementation.

- Web `lib/config.ts` keeps only its zustand/React wiring and re-exports the rest — **zero call-site changes** across its 21 importers.
- `types/engine.ts` / `lib/loopGrid.ts` re-export the config-semantic enums (DCW / RCFG / time-signature / loop-grid), now single-sourced in the SDK. UI-only metadata (`SLIDER_META`, label maps, keyscales) stays in web.
- `applyConfig` / `captureRtmgConfig` / `buildConfig` become thin wrappers over the SDK's pure cores (`applyConfigToState` / `captureConfigFromState` / `rtmgConfigToSessionConfig`).
- **preserve-unknown-on-write**: unknown top-level keys are stashed on a non-enumerable carrier through `mergeConfig` and re-emitted by `serializeConfig`, so a config a client doesn't fully model round-trips without loss. The web Export path serializes through it.

## Decisions

- `schema_version` is already the existing `version` field (added in #266) — not duplicated.
- The config enums can't be sourced from the generated wire types (`time_signature` is bare `string`; `VALID_TIME_SIGNATURES` is ints in `constants.py`), so the SDK is the single source and web re-exports them.
- **`resolveLoraCapForSource` now treats the static `max_concurrent_loras: 0` as uncapped** (was a hard cap of 0). The old web code used `fallback >= 0`, which contradicted the field's own doc ("non-positive means no cap"); the SDK uses `fallback > 0`, matching the documented intent (codified in `configRoundTrip.test.ts`). Only the bare top-level `0` value changes meaning — unset/`null`/negative, positive caps, and `max_concurrent_loras_tiers` (incl. a tier `cap: 0`) are all unaffected. No known deployment sets a bare `0` to block LoRAs, but worth a grep of deployed `config.json` files to confirm.

## Dependency

**Depends on #266** (`refactor/shareable-config`). This PR is stacked on that branch and targets it as its base — retarget to `main` once #266 merges.

## Verification

- `tsc --noEmit` clean; **51/51** web unit tests (incl. new `configRoundTrip.test.ts`); **18/18** Python drift guards (`test_client_sdk` + `test_wire_contract`); SDK `dist/` rebuilt.
- The `test:replay` tier fails, but **pre-existing** on the base branch (canonical recordings predate the `client_time` / `slice_bytes_rx` params telemetry from #260) — confirmed the identical failure with these changes stashed.


leszko (approved):
Commit a small change. LGTM

PR #268, opened 2026-06-16
Update project page with hand-control demo

ryanontheinside:
## Summary
- Replace the prompt-blend sample card with the new bird hand-control demo video from R2
- Reframe the card copy around DEMON supporting custom MediaPipe/performance interfaces
- Add a poster still generated from the new video

## Checks
- Served docs locally at http://127.0.0.1:8080/ and confirmed the new video URL/copy are present and the old prompt-blend sample URL/title are absent
- Verified the generated poster image visually


viborc (approved):
LGTM ✅

Reviewed locally (served` docs/` and compared against the live page). Checked each claim:

- New video is live.` bird_demo_hands.mp4` on the R2 bucket returns 200 (23.8 MB, uploaded same day as the commit).
- The poster is valid.` poster-bird-demo-hands.jpg`, 1920×1080, matches the video.
- Old card fully removed. No leftover `prompt-blend` references except the unrelated LLM-agent caption that mentions "prompt-blend updates" as a control parameter, which is correct and not a stray.
- Copy is accurate. The "separate MediaPipe hand-tracking frontend maps gestures onto DEMON's live controls" line checks out. The `apps/summon` demo in `demon-example-apps` mounts as an external static demo via the shared `/sdk/` bundle and uses MediaPipe hand tracking as a control surface (see `demos/realtime_motion_graph_web/README.md`).
- Base branch is correct. Targets `docs/project-page`, which is the GitHub Pages publishing branch (`/docs` path), so merging publishes straight to the live page.

There's one non-blocking nit: `docs/assets/img/poster-prompt-blend.jpg` is now orphaned. This PR removes its only reference but leaves the file. Fine to drop here or clean up later.

Your choice.

PR #269, opened 2026-06-16
feat(client): native (Node/CJS) targets + public write_audio sender

ryanontheinside:
## What

Enabling work in `@demon/client` so the native hosts (the Max for Live bridge now; the VST next) can adopt the SDK as the **single wire-contract source of truth**, killing the hand-reimplemented protocol copies.

- **`sendWriteAudio`** — public sender for the `write_audio` command (the input primitive behind every M4L feed mode and the VST live splice). Field set (`at_s`/`mix`/`repeat`/`source_epoch`/`refresh_timbre`) derived from the registry in `protocol.py`; `dropIfBusy` is the client-side backpressure gate.
- **`audio_written` / `audio_write_failed`** — dispatch these registered events instead of letting them fall through to the generic `json` event; completes the `write_audio` ack surface.
- **`get sliceEpoch()`** — expose the source-generation counter consumers need to tag local buffers and pin `write_audio` sends.
- **`WebSocketConstructor` option** — inject a WebSocket impl (the bridge passes `ws`); the `OPEN` ready-state constant is resolved from it so the senders never read a global `WebSocket` (Node-for-Max has none reliably).
- **`node.ts` + CJS build target** — a single self-contained `dist/demon-client.node.cjs` (fzstd inlined, no browser-audio code) for `require()`-based hosts. Browser `dist/demon-client.js` regenerated.

## Scope / dependency

- **Stacked on #267** (base is `refactor/client-sdk-config`). Reviewed separately so #267 stays scoped; merge after #267.
- **Blocks the M4L bridge PR** in `daydreamlive/rtmg-vst` (which vendors this branch's `dist/demon-client.node.cjs`).

## Verification

Additive surface. Web app `tsc --noEmit` passes; SDK boundary guard + wire-contract drift guard pass (18/18). The new senders/events/WS-injection are exercised end-to-end by the bridge's mock-server suite in the dependent PR (25/25).

### DIFF
--- a/demos/realtime_motion_graph_web/web/tests/unit/sliceEpoch.test.ts
+++ b/demos/realtime_motion_graph_web/web/tests/unit/sliceEpoch.test.ts
@@ -122,6 +122,45 @@ describe("RemoteBackend slice epochs", () => {
     expect(slices[slices.length - 1].epoch).toBe(3);
   });
 
+  it("adopts the server's source_epoch on swap_ready instead of a blind ++", async () => {
+    // The write_audio pin must track the server's authoritative source
+    // generation. When swap_ready carries source_epoch, the client takes
+    // it verbatim — so a missed or duplicated swap_ready can't silently
+    // desync the local counter from the source the server actually holds.
+    const { remote, ws } = await openSession();
+    ws.emitMessage(
+      JSON.stringify({
+        type: "swap_ready",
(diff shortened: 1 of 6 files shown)


leszko (approved):
Pushed a small fix. Other than that, LGTM

PR #291, opened 2026-06-18
feat(client-sdk): C++ config + controls contract emitter

ryanontheinside:
C++ config/controls contract emitter for rtmg-vst (Plan 06). Drift guard green; MSVC C++17 /WX.


leszko (approved):
Pushed a small fix, then LGTM

PR #298, opened 2026-07-01
perf(sa3): skip wasted random init on model load (~30s -> ~12s)

leszko:
SA3's loader builds the model with random weights on CPU (~20s), then immediately overwrites every parameter from the checkpoint — so the init is pure wasted work.

`skip_param_init()` (in `acestep.engine.sa3_helpers`) no-ops the random fills during construction only; `load_state_dict` and deterministic buffers are untouched. `load_local_model` wraps `load_diffusion_cond` in it.

- **Bit-identical** to the un-skipped load across every param and buffer of the medium checkpoint.
- Cold `SA3Context` load drops **~30s → ~12s** on a 5090.
- `tests/unit/test_sa3_fastload.py` locks the contract with a toy module (no GPU/weights in CI).

### DIFF
--- a/acestep/engine/sa3_helpers.py
+++ b/acestep/engine/sa3_helpers.py
@@ -12,6 +12,7 @@ with a local checkout.
 
 from __future__ import annotations
 
+import contextlib
 import os
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


ryanontheinside (approved):
sounds good to me

PR #305, opened 2026-07-13
feat(provenance): local session logs, C2PA manifests, pod-side slice hashing (P1)

hthillman:
## Local provenance: session logs, C2PA manifests, pod-side slice hashing (P1)

Engine half of the provenance system: every DEMON session produces a schema-stable local event log, file writers embed C2PA manifests, and pods report hash-chained events + per-slice hashes to the provenance ledger.

### What's here
- **`acestep/provenance/`**: §2.2-schema JSONL session logs (epoch-ms, namespaced event types, `stream:"local"`); local ES256 mini-CA ("DEMON (local, self-signed)"); c2pa-python manifest embedding in `write_track_wav`/`write_stem_wavs`; ledger client speaking the real wire contract (batched seq-contiguous envelopes, Bearer auth, `slice.pod_hash` events, receipt parsing). Optional extra: `provenance` (`c2pa-python>=0.36`). Fully no-op when the extra/env is absent.
- **Pod-side slice hashing**: SHA-256 over the exact pre-zstd interleaved f16 delta bytes at `SliceCodec.encode` — the same bytes the plugin hashes on receipt, making the pod/client dual-hash cross-check real.
- **Honest labeling**: uploaded human audio (the only assets these writers touch) is marked `digitalCapture` — never `compositeWithTrainedAlgorithmicMedia`. Do-not-train assertion on by default. Manifests bind to an explicit session id (no "latest session" races).

### Testing
17 tests green: `uv run --no-sync python -m pytest tests/unit/test_provenance_manifest.py tests/unit/test_provenance_session_log.py --noconftest -q` — includes a real c2pa round-trip (write WAV → read manifest → assert digitalSourceType, do-not-train, com.daydream.session) and HTTP-server-backed ledger client tests. Degrade path verified with c2pa absent.

### Cross-repo
Spec: `demon-provenance` docs. Server: https://github.com/livepeer/pipelines/pull/2709 · Plugin: https://github.com/daydreamlive/rtmg-vst/pull/162

### DIFF
--- a/acestep/provenance/__init__.py
+++ b/acestep/provenance/__init__.py
@@ -12,6 +12,9 @@ demon-provenance spec, 02-architecture.md §2/§3/§5/§7):
   embedding on the track-asset WAV writers via ``c2pa-python``.
 - :mod:`~acestep.provenance.ledger_client` — thin event/slice-hash
   POST client, a complete no-op until ``DEMON_LEDGER_URL`` is set.
+- :mod:`~acestep.provenance.receipts` — client-side verification of
+  the ledger's signed receipts (spec 06 §2.4: key-cert chain, JCS +
+  chain-head recomputation, Ed25519 framing checks).
 
 Everything here degrades to a no-op (one logged warning, never an
 exception on a write path) when the optional ``provenance`` extra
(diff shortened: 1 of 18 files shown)


hthillman:
Added `7f9ded4`: client-side receipt verification (06-cloud-api §2.4 step 5) — JCS canonicalizer + chain mirror + Ed25519 receipt/key-cert verification, fail-open with verified/unverified counters. Cross-implementation golden vector: the Python chain heads byte-match the server's `chain.test.ts` fixture exactly, fuzz-validated against `jcs.ts` semantics (9k cases, zero mismatches). 33 tests green.

PR #329, opened 2026-08-23
feat(provenance): Level 1 action log — server-side session recipe logging

leszko:
## What

The **Level 1 "Log Actions"** cut of the provenance work ([design doc](https://app.notion.com/p/livepeer/Daydream-Provenance-3c30a34856878025b56fc1d04f69e116), per the team discussion to start light): record every user operation of a live session — actions, parameter changes, prompts, input-source hashes, models used — to a local JSONL log and, when configured, to the cloud action log.

This is the lightweight **alternative** to the full provenance branch (#305). Compared to it, this PR deliberately drops: audio slice hashing + canvas fingerprints, receipt verification, C2PA manifests on written WAVs, and local signing keys. The pod makes **no cryptographic claims** at this tier — the log answers "what happened in this session", not "prove this WAV came from it".

Pairs with livepeer/pipelines#2758 (the Level-1 ledger service). No rtmg-vst changes are needed at this tier — the log is entirely server-side.

## What ships

- **`acestep/provenance/session_log.py`** — event-bus tap writing the shared `{stream, seq, type, ts, payload}` envelope per session (`session.config` opener, `action.prompt` / `action.param` / `action.lora` / `action.seed_audio` / `action.transport`, `input.source`, `session.note` seal). Params are diffed + rate-limited so the ~125 Hz knob stream never floods the log. Input sources are recorded as `input.source` events: the buffer fingerprint of what was actually consumed (covers client uploads that never exist as pod files) plus, when a cached source file exists, its plain `sha256` (comparable to `shasum -a 256 <file>`). Decoded slices are counted, never hashed.
- **`acestep/provenance/ledger_client.py`** — batched fire-and-forget POST client for the pod event stream. Contiguous `seq` assigned at flush time, so best-effort queue drops never trip the server's gap check. Dev-only broker emulation (`DEMON_LEDGER_INTERNAL_SECRET`) mints the session via the internal endpoint for E2E testing; production pods receive the `dlt_pod_…` token from the broker.
- **`ParamsEcho.origin`** (`"external"` / `"primary"`): PRIMARY knob changes are now published on the bus for the provenance tap — the performer's moves are the authorship record — and never echoed back to the client (which would fight its own UI tween). Transports forward only external echoes; additive, default keeps today's behavior.
- **registry / ws_adapter wiring** — `SessionHandle` gains optional `bus` + `provenance_meta`; `register()`/`unregister()` attach/detach the tap. The adapter hands over the checkpoint, fixture name, and input fingerprints — never raw server filesystem paths.
- **Fix**: the first knob change of a session could be silently rate-limited away on platforms where `time.monotonic()` starts near zero (the window was compared against an initial `0.0`).

Everything is fail-open: provenance can never block session registration, the bus drainer, or the downlink.

## Tests

10 unit tests (pure Python, no GPU, no new deps): envelope schema + lifecycle, input-source recording, params diff + rate-limit, primary-echo handling, registry inertness without a bus, ledger client wire shape / auth / contiguous seq / dev bootstrap / fail-open.

### DIFF
--- a/acestep/provenance/__init__.py
+++ b/acestep/provenance/__init__.py
@@ -1,0 +1,46 @@
+"""Local provenance: the Level-1 session action log.
+
+The lightweight cut of the DEMON provenance architecture (see the
+demon-provenance design doc, "Level 1: Log Actions"): record what
+happened in a session — actions, parameter changes, prompts,
+input-source hashes, models — locally and to the cloud action log.
+Deliberately NO cryptographic machinery here (no slice hashes, no
+receipts, no C2PA); that is the follow-up tier.
+
+- :mod:`~acestep.provenance.session_log` — local JSONL session logs
+  using the same event schema as the cloud action log, tapped off the
+  streaming event bus.
+- :mod:`~acestep.provenance.ledger_client` — thin batched event POST
+  client, a complete no-op until ``DEMON_LEDGER_URL`` is set.
+
+Everything here degrades to a no-op (one logged warning, never an
+exception on a write path).
+"""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+_ENV_PROVENANCE_DIR = "ACESTEP_PROVENANCE_DIR"
+_DEFAULT_PROVENANCE_DIR = os.path.join(
+    os.path.expanduser("~"), ".daydream-scope", "provenance",
+)
+
+
+def provenance_dir() -> Path:
+    """Root directory for local provenance state (session logs).
+
+    Resolution order mirrors :func:`acestep.paths.models_dir`:
+        1. ACESTEP_PROVENANCE_DIR environment variable
+        2. ~/.daydream-scope/provenance
+
+    Deliberately NOT memoized: tests point it at a tmp dir per case,
+    and it is read only on session start, never per tick.
+    """
+    return Path(os.environ.get(_ENV_PROVENANCE_DIR, _DEFAULT_PROVENANCE_DIR))
+
+
+def session_logs_dir() -> Path:
+    """Directory holding per-session JSONL event logs."""
+    return provenance_dir() / "sessions"
--- a/acestep/provenance/ledger_client.py
+++ b/acestep/provenance/ledger_client.py
@@ -1,0 +1,316 @@
+"""Thin client for the Provenance Action Log service.
+
+Best-effort background reporter for one session's event stream. It
+batches events and POSTs them to::
+
+    POST {DEMON_LEDGER_URL}/sessions/{sessionId}/events
+    Authorization: Bearer {DEMON_LEDGER_TOKEN}
+    { "events": [ {seq, type, ts, ppq?, payload}, ... ] }
+
+This is the Level-1 "recipe log" tier: the service acknowledges with a
+plain ``{event_count}`` — no hash chains, no signed receipts.
+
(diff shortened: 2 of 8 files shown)


leszko:
Note: the paired API PR (livepeer/pipelines#2758) dropped the 90-day retention sweep and pin/unpin — action logs are now kept until the user deletes them. No change needed on this side; the pod just reports events.

leszko:
Pushed 69c5f78: styles are now identified by the **SHA-256 of the adapter's weight file**, not the catalog id.

Why: the catalog id is the filename stem. It is chosen by whoever materialized the file, it collides across users, and it joins to nothing — so "which style ran" could not be answered off-pod. The weight hash is what the registry's `lora_artifacts.sha256` already stores, so a logged style is one join away from the row that published it, and from its owner. That is the artist-payout question (paired API PR: livepeer/pipelines#2758).

`action.lora` now carries, per enabled style: `sha256`, display `name`, catalog `id`, `trigger_word`, `size_bytes`, weight `family`, `base_model`, `strength`.

Two things worth a reviewer's eye:

- **Hashing never sits on a hot path.** A miss is a 10–400 MB read. `lora_identity` memoizes on `(path, size, mtime_ns)` — the same key the metadata sidecar loader uses — so a style toggled repeatedly in one session hashes once, and `pending()` lets the tap ask whether the answer is cached. When it isn't, the *whole event* is deferred to a daemon thread rather than published without a hash (a hashless style is one nobody can be paid for). Same pattern as the existing off-thread input-source hash.
- **The event timestamp is captured before that wait**, and `_record` now accepts it, so a deferred event still says when the change happened. The seal reconstructs per-style enabled intervals from these timestamps, so getting this wrong would mis-bill a style.

`action.lora` is also now the single owner of style state: it fires for the set the session opened with — even when empty, so the seal can distinguish "started with no style" from "this pod never reported styles" — and again on every catalog change. `session.config.loras` stays as a mirror of the WS handshake, and the API ignores it whenever `action.lora` is present.

Fails open like the rest of the module: an unreadable weight file yields `sha256: null` and the event still names the style.

Tests: 13 pass (3 new — deferred hashing, mid-session style changes, cache identity).

leszko:
Superseded for the current goals by **daydreamlive/daydream-plugins#317**, which sends the same information — parameter moves, prompts, and per-Style usage with durations — straight from the plugin into PostHog, with no backend.

Per the Discord thread: we're not doing provenance or C2PA now and not monetising Styles yet, so the two live goals ("know what users are doing", "know which Styles get used") are analytics questions. PostHog already receives ~100 `vst_*` events from the plugin and forwards everything to R2, so the marginal cost was three events rather than a schema, an ingestion API, pod tokens and an admin console. It also sidesteps the blocker on this design: nothing forwards the pod its ledger token in production yet, so this tier would have recorded nothing there.

Worth keeping this branch around rather than deleting it. It is still the right shape for two things PostHog can't do:

- **Payouts.** Client-reported numbers are forgeable once money is attached, PostHog has retention windows and lossy ingestion, and there's no transactional join to the Style registry. When Styles are licensed, the usage figure an artist is shown and the figure they're paid on must be the same durable number — that's `provenance_lora_usage` plus the seal rollup from the paired API PR, and it does *not* need the full event log.
- Anything cryptographic (slice hashes, provable WAV, C2PA — Levels 2–6 of the design doc), which needs the timeline this PR produces.

No action needed here; flagging so nobody reviews it as live work.

PR #333, opened 2026-08-28
feat(enhance): optional local prompt-enhancer backend, and a variations grid

gioelecerati:
## What

Two additions to the prompt-enhancement surface, both optional and both off by default.

**A second provider for `/api/enhance`.** Today it can only ask a hosted LLM: that needs an API key, a network round trip, and a per-call fee for answers that repeat constantly within a session. This adds a small seq2seq model (t5-small class), fine-tuned on that LLM's own output, running in-process — roughly a tenth of the latency, no key, deterministic.

**`/api/variations`**, which answers *"what else is near this prompt?"* — one coordinate per call.

## Configuration

| variable | meaning |
|---|---|
| `DEMON_ENHANCER_PROVIDER` | `hosted` (**default**, identical to today) / `local` |
| `DEMON_ENHANCER_REPO` | HuggingFace repo to fetch the checkpoint from |
| `DEMON_ENHANCER_DIR` | local checkpoint path, for images that stage it themselves |
| `DEMON_ENHANCER_DEVICE` | `cpu` (**default** — the GPU is running a realtime audio pipeline) / `cuda` / `auto` |
| `HF_TOKEN` | read token, when the repo is private |

`/api/enhance?provider=` overrides per request, but may only narrow: a client cannot arm `local` on a pod configured `hosted`. `provider` picks *who answers*; the existing `backend` picks *which prompt policy* — orthogonal, hence two parameters.

## Why it is safe to merge before anyone uses it

- The default is `hosted`. An existing deployment sets nothing and behaves exactly as before — verified: with no API key the default still declines and echoes the caller's text unchanged with `ok=false`.
- The model is narrow. It was trained on structured prompt text and degrades on free-form input, so it returns `""` for anything it cannot handle and the hosted policy answers instead. A missing checkpoint takes the same path.
- The checkpoint is **not in the repository** and the download is optional. A failed or skipped fetch costs the feature, never the pod.

## `/api/variations`

Distance from the anchor collapses to an integer (`stop`: how many leading tokens stay fixed, bounded to `MAX_FREE` of the line), and variety at a given distance is `LANES` discrete sampling streams (`lane`). Both are deterministic, so a client can cache every coordinate it has seen and travelling out and back is lossless.

```
GET /api/variations?prompt=<anchor>&stop=<0..15>&lane=<0..11>
-> {"text": str, "stop": int, "lane": int, "ok": true}
```

Answering the whole neighbourhood in one call was tried and **withdrawn**: a grid holds the generation lock for its entire duration (seconds), during which nothing else on the endpoint can answer — one coordinate is ~0.14 s on a 5090. Concurrent callers get an immediate `503 {"busy": true}` with `Retry-After` rather than queueing in front of the live audio thread.

Two implementation choices that are load-bearing rather than incidental:

- **Batched along the lane axis, even for a single coordinate.** Sampling consumes the random stream per batch, so a narrower batch would give a lane a different draw than the same lane computed alongside its siblings — the answer would change depending on how it was asked for. Every request computes the full lane batch for its stop and indexes it.
- **One seed for the whole travel** (`crc32` of the anchor — not `hash()`, which is salted per interpreter), reset per stop, so each distance starts from an identical RNG state and differs only in how much of the line was freed. Seeding per stop is equally reproducible, but makes every distance an independent draw — measured, that roughly doubled the edit distance between adjacent steps, turning travel into a shuffle you can slide back to.

## Verified

Against the real checkpoint:

- stop 0 returns the anchor **exactly**, in every lane — travelling away and back is lossless
- all 12 lanes distinct at maximum distance
- byte-identical output across repeated calls and process restarts
- default provider unchanged with no API key; an unknown provider value falls back to `hosted`

Plus a checkpoint-free unit suite (`tests/unit/test_prompt_variations.py`) covering coordinate clamping, prefix monotonicity, seed stability, the non-blocking busy gate (under a deadlock watchdog), query routing, and download-failure residue — no torch, no network, so it gates in CI.

## Post-review updates (2026-08-31)

- Rebased onto `main` (the SOUND/SCENE prompt-policy section landed separately as #334 and deduplicated cleanly).
- Fixed an `UnboundLocalError` in `download_prompt_enhancer_model`: `created` was assigned after the lazy `huggingface_hub` import inside the same `try`, so on an image without that package the `except` crashed instead of degrading. The covering test now forces the `ImportError` and a second test stubs `snapshot_download` — neither touches the network.
- `resolve_provider` is public now that `server.py` imports it; the `auto` provider value was dropped entirely (the fallback design made it an unobservable synonym for `local` — an unknown value still falls back safely to `hosted`); dead `available()` helper removed; stale whole-grid comments brought in line with the one-coordinate contract.

### DIFF
--- a/acestep/model_downloader.py
+++ b/acestep/model_downloader.py
@@ -202,6 +202,14 @@ PROMPT_ENHANCER_MODEL_NAME = "prompt-enhancer"
 PROMPT_ENHANCER_DEFAULT_REPO = "daydreamlive/t5-small-enhancer"
 #: The files transformers needs to load it; anything else in the repo is
 #: ignored so a card or extra format does not bloat every pod.
+#: What must be present for the checkpoint to load at all. A t5 fine-tune may
+#: legitimately ship either a fast tokenizer (tokenizer.json) or a slow one
+#: (spiece.model), so neither is required on its own -- see the check below.
+PROMPT_ENHANCER_REQUIRED = [
+    "config.json",
+    "model.safetensors",
+    "tokenizer_config.json",
+]
 PROMPT_ENHANCER_ALLOW = [
     "config.json",
     "generation_config.json",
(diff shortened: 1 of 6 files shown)


leszko (approved):
LGTM

PR #335, opened 2026-08-29
fix(sa3): drop old-anchor latents on source swap

ryanontheinside:
After a source swap the SA3 backend kept rendering covers of the previous source: the cached latent (gap-fill, DiT-pause reuse) and the slots still in flight, which emerge over the next pipeline depth. The old song played over the new one for a second or more after every swap.

- `handle_swap_source` drops `_last_result_latent` / `_current_result` (ACE already does this in `sync_source`).
- `_generate` discards a finished latent whose request was built on a previous anchor (`x0_target` identity), logged as `sa3_gen_discarded reason=source_swapped`.

Until the first new-anchor slot completes the runner writes nothing and the client plays the swapped-in source, same as session create. Observed live: 5 discards per swap at the default depth, then clean output.

Tests: new `test_handle_swap_source_discards_covers_of_the_old_source`; `tests/unit` 553 passed, the 4 `test_lora_facade_session` failures are pre-existing on main.

### DIFF
--- a/acestep/streaming/sa3_backend.py
+++ b/acestep/streaming/sa3_backend.py
@@ -857,11 +857,17 @@ class SA3Backend(DiffusionBackend):
         taps are covers of the OLD source; blending them into the new
         anchor would smear the previous song across the swap).
 
-        In-flight pipeline slots finish on the old anchor — the swap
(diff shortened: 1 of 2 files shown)


leszko (approved):
LGTM

PR #340, opened 2026-09-02
feat(sa3): sa3_sigmas — an explicit per-step noise schedule

gioelecerati:
The streaming analogue of StreamDiffusion's `t_index_list`, asked for from the plugin: *"can I directly have `sigmas = [1.0, 0.85, 0.65, 0.3, ...]` on the vst?"*

## What existed

SA3's schedule was a three-parameter family — step count, `sigma_max` (`sa3_denoise`) and the `sa3_shift` warp over the checkpoint's stock curve. You could cluster steps toward noise or toward clean; you could not *place* them. Two steps at two chosen levels had no expression.

## The knob

`sa3_sigmas`: a list of per-step levels in (0, 1], strictly decreasing, `1.0` meaning "start from `sa3_denoise`", the final 0 implied. `[1.0, 0.85, 0.65, 0.3]` is four steps at exactly those levels.

- **Its length is the step count** and overrides `steps_override` while set. A length change goes through the existing pipeline rebuild — schedules are `(steps+1,)`-shaped and the per-denoise cache would otherwise serve the wrong size.
- **SHIFT still composes on top**, so it becomes a macro over a hand-placed schedule rather than a dead knob.
- **Not a `KnobSpec`, deliberately.** The registry types scalars only and would coerce a list to NaN. It rides the params channel as a pass-through key (`coerce_knob_values` leaves unregistered keys verbatim; `KnobState` keeps them) and is validated in the backend by `parse_sigmas`, which accepts a list or the comma/space text a person types.
- **Anything invalid is ignored, not applied** — including a half-typed list — so the previous schedule keeps playing until the text makes sense. The plugin mirrors the same parser, so what its field accepts is what the pod applies.
- The adapter refuses a list whose length disagrees with `infer_steps` rather than truncating a schedule someone placed by hand.

TensorRT is unaffected: `t` is a runtime input, and step count is just how many times the DiT is called.

## Tests (52 pass, up from 44)

Override replaces the stock schedule and scales by strength · SHIFT composes with pinned endpoints · clearing restores the builder exactly · length mismatch fails loudly · `parse_sigmas` accepts only sane input (rejects `1.0, 0.5, 0.5`, `0.5, 1.0`, `1.0, 0.0`, `1.2`, NaN, 17 entries) · a list applies in place and invalidates the cache · a longer list rebuilds the pipeline and survives it · an unchanged list does not churn the cache · invalid text falls back to `steps_override`.

Run on CPU torch here (`tests/unit/test_sa3_adapter.py`, `test_sa3_backend.py`, `test_knob_homonyms.py`).

## Pairs with

The rtmg-vst side adds a SIGMAS field on the SA3 deck that sends this key (separate PR). **Needs a pod re-bake to reach the fleet.**

## Caveat worth saying out loud

The serving model is the ARC distillation, trained around its stock schedule. `sa3_shift` being bounded to 0.25–4 suggests there's a cliff somewhere; a hand-placed list can walk off it. This lands the lever; finding the usable range is the listening that comes next.

https://claude.ai/code/session_015jmVRt5vLoU9b7WE5ZiqKp

### DIFF
--- a/acestep/streaming/sa3_backend.py
+++ b/acestep/streaming/sa3_backend.py
@@ -29,6 +29,14 @@ Control surface (everything else off, capability-gated):
 * ``sa3_shift`` — relative schedule warp on top of the checkpoint's
   dist_shift (``SA3Adapter.shift_alpha``); changes invalidate the
   pipeline's per-denoise schedule cache.
+* ``sa3_sigmas`` — an EXPLICIT per-step schedule, the streaming analogue
+  of StreamDiffusion's ``t_index_list``: a list of noise levels in
+  (0, 1], strictly decreasing, 1.0 meaning "start from ``sa3_denoise``"
+  and the final 0 implied. Its length IS the step count, so it overrides
+  ``steps_override`` while set. Not a :class:`KnobSpec` on purpose —
+  the registry only types scalars and would coerce a list to NaN — so
+  it rides the params channel as a pass-through key, validated here by
+  :func:`parse_sigmas`; anything invalid is ignored rather than applied.
 * ``x0_target`` / ``feedback`` / ``feedback_depth`` — taken FROM the
   shared registry (identical semantics to ACE, solver-level latent
   mechanics that are family-agnostic): the source-lock morph toward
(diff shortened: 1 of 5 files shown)


ryanontheinside:
SUPER sick dude

ryanontheinside (approved):

acestep/streaming/sa3_backend.py
```diff
@@ -1274,6 +1319,14 @@ def _prepare_tick(self, knobs: dict, ctx: TickContext) -> dict:
                 self.adapter.shift_alpha = shift
                 self.pipeline.invalidate_schedule_cache()
 
+        # Explicit schedule: same hot-apply + cache-invalidate contract as
+        # the warp. Compared as lists so an unchanged list is a no-op tick.
+        sigmas = parse_sigmas(knobs.get("sa3_sigmas"))
```
ryanontheinside, line 1324: Could invalid input retain the last valid schedule? Today a partial edit like 1.0, 0. clears it; for live control, a short lerp between valid same-length schedules would also prevent jumps. Convex interpolation preserves monotonicity.

-Codex

acestep/streaming/sa3_backend.py
```diff
@@ -1311,7 +1364,12 @@ def _prepare_tick(self, knobs: dict, ctx: TickContext) -> dict:
         return {
             "denoise": float(knobs.get("sa3_denoise", 1.0)),
             "seed": int(knobs.get("seed", self._default_seed)),
-            "steps": int(knobs.get("steps_override", self._steps)),
+            # The list's length is the step count while it is set; the
+            # (steps+1,)-shaped schedule cache makes any other value wrong.
+            "steps": (
```
ryanontheinside, line 1369: Could rebuild_imminent() use this same effective step count? A sigma-length change rebuilds the pipeline, but the current pre-stall check only sees steps_override, so it misses pre-coverage.

-Codex

